PR #1, opened 2015-04-09
add option to poll stats for all nodes in the cluster

dan-cleinmark:
Tested on a 6 node cluster (1 master, 5 data nodes).  Might cause timeouts on very large clusters.

    $ time curl -s "http://localhost:9200/_nodes/_local/stats" &> /dev/null
    real	0m0.011s
    user	0m0.000s
    sys	0m0.004s

    $ time curl -s "http://localhost:9200/_nodes/stats" &> /dev/null
    real	0m2.164s
    user	0m0.008s
    sys	0m0.004s

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,12 @@ make
 ./elasticsearch_exporter --help
 ```
 
-* __`es.server`:__ Address (host and port) of the Elasticsearch node we should
+* __`es.uri`:__ Address (host and port) of the Elasticsearch node we should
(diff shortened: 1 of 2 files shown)


ewr:
@dan-cleinmark: So sorry about not seeing these before. Apparently I wasn't watching my repo, and therefore didn't get notified. My fault.

We're having a bigger cleanup discussion in #4, and I want to make sure that your submissions get pulled in as part of that effort.

PR #2, opened 2015-04-09
specify expected json fields within indices response

dan-cleinmark:
filter_cache/* was returning 0 in the exporter without an explicit field mapping, updated other fields for consistency.


ewr:
This was fixed in c73014f94f31addee1dffe1672b633fcf4ab8dcc, which just went to master.

PR #3, opened 2015-04-09
export OS CPU & disk stats

dan-cleinmark:


matthiasr:
Why not just use the [node exporter](https://github.com/prometheus/node_exporter)?

dan-cleinmark:
@matthiasr that's definitely an option.  This came on the back of [this PR](https://github.com/ewr/elasticsearch_exporter/pull/1) to get just enough stats from every node in the cluster through a single exporter.

PR #4, opened 2015-06-16
Cleanup and best pratices

discordianfish:
Hi,

we would like to use the elasticsearch_exporter in our infrastructure. Since I'm also a prometheus contributor and wrote a few exporters, I thought I do frist some cleanup and changes to better follow prometheus and go's best practices.

Beside some minor (but IMO important) changes to mainly naming, the probably biggest change is the removable of the node label for the metrics.

Since the exporter returns stats for the local node only, it's not necessary to set this as a label since prometheus will include the instance label to identify the node. If a custom node name is
necessary, this can be specified in the prometheus target configuration:

http://prometheus.io/docs/operating/configuration/#target-groups-target_group

What do you think?


elasticsearch_exporter.go
```diff
@@ -267,52 +26,52 @@ type VecInfo struct {
...
-		"indices_docs_count":                        "Count of documents on this node",
-		"indices_docs_deleted":                      "Count of deleted documents on this node",
-		"indices_store_size_in_bytes":               "Size of stored index data in bytes",
-		"indices_segments_memory_in_bytes":          "Memory size of segments in bytes",
-		"jvm_mem_heap_committed_in_bytes":           "JVM heap memory committed",
-		"jvm_mem_heap_used_in_bytes":                "JVM heap memory used",
-		"jvm_mem_heap_max_in_bytes":                 "JVM heap memory max",
-		"jvm_mem_non_heap_committed_in_bytes":       "JVM non-heap memory committed",
-		"jvm_mem_non_heap_used_in_bytes":            "JVM non-heap memory used",
+		"indices_fielddata_evictions":            "Evictions from field data",
```
discordianfish: @ewr Are you sure this is a gauge? I couldn't find any documentation about whether this is a gauge or counter. It looks like a counter to me. Same for the filter_cache_evictions below.

elasticsearch_exporter.go
```diff
@@ -335,41 +94,43 @@ type Exporter struct {
...
+	gauges := make(map[string]*prometheus.GaugeVec, len(gaugeMetrics)+len(gaugeVecMetrics))
 
 	for name, info := range counterVecMetrics {
-		log.Printf("Registering %s", name)
+		log.Println("Registering", name)
 		counters[name] = prometheus.NewCounterVec(prometheus.CounterOpts{
 			Namespace: namespace,
 			Name:      name,
 			Help:      info.help,
-		}, append([]string{"cluster", "node"}, info.labels...))
```
brian-brazil: The cluster label should also go away. If there's such a thing it should be provided on the prometheus end.

elasticsearch_exporter.go
```diff
@@ -335,41 +94,43 @@ type Exporter struct {
...
+		}, append([]string{"cluster"}, info.labels...))
 	}
 
 	for name, info := range gaugeVecMetrics {
-		log.Printf("Registering %s", name)
+		log.Println("Registering", name)
```
brian-brazil: These are debug logs at best

elasticsearch_exporter.go
```diff
@@ -434,82 +195,81 @@ func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
...
+	}
 
-	for node, stats := range all_stats.Nodes {
-		log.Printf("Processing node %v", node)
+	for _, stats := range allStats.Nodes {
 		// GC Stats
 		for collector, gcstats := range stats.JVM.GC.Collectors {
-			e.counters["jvm_gc_collection_count"].WithLabelValues(all_stats.ClusterName, stats.Name, collector).Set(float64(gcstats.CollectionCount))
-			e.counters["jvm_gc_collection_time_in_millis"].WithLabelValues(all_stats.ClusterName, stats.Name, collector).Set(float64(gcstats.CollectionTime))
+			e.counters["jvm_gc_collections"].WithLabelValues(allStats.ClusterName, collector).Set(float64(gcstats.CollectionCount))
```
brian-brazil: jvm_gc_collection_seconds_count

elasticsearch_exporter.go
```diff
@@ -434,82 +195,81 @@ func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
...
+			e.counters["jvm_gc_collections_time_ms"].WithLabelValues(allStats.ClusterName, collector).Set(float64(gcstats.CollectionTime))
```
brian-brazil: This should be jvm_gc_collection_seconds_sum, with appropriate time conversion

elasticsearch_exporter.go
```diff
@@ -434,82 +195,81 @@ func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
...
 		}
 
 		// JVM Memory Stats
-		e.gauges["jvm_mem_heap_committed_in_bytes"].WithLabelValues(all_stats.ClusterName, stats.Name).Set(float64(stats.JVM.Mem.HeapCommitted))
-		e.gauges["jvm_mem_heap_used_in_bytes"].WithLabelValues(all_stats.ClusterName, stats.Name).Set(float64(stats.JVM.Mem.HeapUsed))
-		e.gauges["jvm_mem_heap_max_in_bytes"].WithLabelValues(all_stats.ClusterName, stats.Name).Set(float64(stats.JVM.Mem.HeapMax))
-		e.gauges["jvm_mem_non_heap_committed_in_bytes"].WithLabelValues(all_stats.ClusterName, stats.Name).Set(float64(stats.JVM.Mem.NonHeapCommitted))
-		e.gauges["jvm_mem_non_heap_used_in_bytes"].WithLabelValues(all_stats.ClusterName, stats.Name).Set(float64(stats.JVM.Mem.NonHeapUsed))
+		e.gauges["jvm_mem_heap_committed_bytes"].WithLabelValues(allStats.ClusterName).Set(float64(stats.JVM.Mem.HeapCommitted))
+		e.gauges["jvm_mem_heap_used_bytes"].WithLabelValues(allStats.ClusterName).Set(float64(stats.JVM.Mem.HeapUsed))
```
brian-brazil: jvm_memory_bytes_used{area="heap"} etc

elasticsearch_exporter.go
```diff
@@ -267,52 +26,52 @@ type VecInfo struct {
...
 	}
 	counterMetrics = map[string]string{
 		"indices_flush_total":                   "Total flushes",
-		"indices_flush_time_in_millis":          "Cumulative flush time",
-		"transport_rx_count":                    "Count of packets received",
-		"transport_rx_size_in_bytes":            "Bytes received",
-		"transport_tx_count":                    "Count of packets sent",
-		"transport_tx_size_in_bytes":            "Bytes sent",
-		"indices_store_throttle_time_in_millis": "Throttle time for index store",
+		"indices_flush_time_ms_total":           "Cumulative flush time in milliseconds",
```
brian-brazil: Naming is inconsistent for ms vs. millis.

I'd suggest seconds everywhere - and be careful with the ES docs. They've lowercased all units, which leads to apparent units such as millibits.

elasticsearch_exporter.go
discordianfish: Hrm.. I assume a node can be part in multiple clusters? If not, I agree that this should be removed as well.

elasticsearch_exporter.go
discordianfish: I thought only summaries should have _count suffix?

ewr:
Sounds good to me. 

Here's my counter-argument for node name, but I'm not set one way or the other: the exporter doesn't have to use an Elasticsearch instance on localhost. That's the default argument value, but not required.

In fact, the README reflects my original intent that one exporter doesn't even have to just pull one node... If you hit `/_nodes/stats` instead of `/_nodes/_local/stats`, you get the same stats for all nodes in the cluster. I didn't end up implementing that, but it's a possible case for how you would want to run it.

elasticsearch_exporter.go
ewr: I bet you're right. Those do sound like counters.

elasticsearch_exporter.go
ewr: An Elasticsearch instance can only be a part of one cluster. A node could run more than one Elasticsearch instance, and therefore be part of more than one cluster, but that's a pretty unlikely case.

You're arguing that this should be the responsibility of the discovery system on the Prometheus config side? I'm not entirely clear on the benefit of that.

ewr:
For discussion: A while back I had tweaked this exporter to roll up all the time and memory metrics into a `time` counter and `memory` gauge. I had apparently never committed this, but just pushed it to get some feedback: 

https://github.com/ewr/elasticsearch_exporter/compare/SummaryMetrics

It made for nice graphing compared to having to list out all the different bytes and seconds metrics, but if it's considered a bad pattern I'm not tied to it.

ewr:
And actually #1 is a PR to implement the all nodes functionality I mentioned. Oops.

elasticsearch_exporter.go
discordianfish: What's inconsistent about that? Re/ seconds, I agree but didn't want to change this since I'm not sure we all can agree on that. I like having every time in seconds. @ewr: What do you think?

elasticsearch_exporter.go
discordianfish: I guess I'll just remove them then.

discordianfish:
@ewr In general I would just run one exporter on every ES instance. If you would run one exporter per cluster, you need somehow make that HA which is more complicated etc.
I know that @brian-brazil agrees and probably has some even better arguments.

That being said, it could of course support both but I wouldn't recommend running it with cluster wide stats and it makes the code more complicated.

Re/ rolling up the metrics, I quote the best practices:
```
As a rule of thumb, either the sum() or the avg() over all dimensions of a given metric should be meaningful (though not necessarily useful). If it is not meaningful, split the data up into multiple metrics. For example, having the capacity of various queues in one metric is good, while mixing the capacity of a queue with the current number of elements in the queue is not.
```
- http://prometheus.io/docs/practices/naming/

ewr:
Yeah, graphing `sum(elasticsearch_memory)` and seeing all the different pieces roll up was neat, as compared to `sum(elasticsearch_jvm_heap_used + elasticsearch_segments_used + ... )`

I'm not sure I was comprehensively pulling everything, but it did seem to pull together likes with likes.

ewr:
Re: cluster vs node—My gut is to leave it in. I wouldn't recommend it for production, like you said, but the code complexity is really just one level in the JSON parsing and a different URL. For someone just getting their feet wet, it seems a nice feature.

elasticsearch_exporter.go
brian-brazil: Yes, this should be on the Prometheus config side.

Attaching labels to targets is the responsibility of the Prometheus server, exporters shouldn't second-guess what the user's label setup is.

elasticsearch_exporter.go
brian-brazil: We're faking a summary here

discordianfish:
@ewr I don't think that adding up all the times for example make sense. It's all time but doesn't make that much sense when all added up, right? Same for the memory. It also shows up in the help string you used: 'Roll up of various' kinda shows that it's not that logical thing you can give a name, right?

Re: cluster vs node: Since it will otherwise duplicate the instance labels (which is ugly and might collide) we really should omit them for the 'one exporter per node' usecase. That's what would make the code more complicated.

brian-brazil:
> If you would run one exporter per cluster, you need somehow make that HA which is more complicated etc.

That's correct, the best way to run this type of exporter is one per node as that keeps the failure domains aligned.

Adding labels such as `node` and `cluster` may clash with other labels the user plans on using (for example they'd be noise in my setup, as I'd be using the standard job and instance labels for that purpose). Unless you've got a type of exporter where there's multiple instances being scraped (which is discouraged), leave the instance and job class labels up to the prometheus configuration.

ewr:
Well, there's a finite amount of work time per node, so the higher the cumulative rate goes, the more likely it is that the cluster won't be very responsive. Added up its "How much time per second did ES spend working?"

ewr:
For the labels, I get that running the exporter on the same instance and running one exporter per node is preferred, but I'm a little uneasy about removing that support entirely. 

For cluster, that's something that you never want to be reporting an incorrect value because your monitoring config is wrong. ES knows its cluster, and it seems like it should be the authority on it. Plus, I don't have a clean way that I would convey that through in my config (which gets generated from Consul service registrations).

brian-brazil:
> Well, there's a finite amount of work time per node, so the higher the cumulative rate goes, the more likely it is that the cluster won't be very responsive. Added up its "How much time per second did ES spend working?"

If the time is partitioned over the labels then this could be okay.

Is cpu time spent by the process exposed anywhere? That's http://docs.oracle.com/javase/7/docs/jre/api/management/extension/com/sun/management/OperatingSystemMXBean.html#getProcessCpuTime()

brian-brazil:
> For cluster, that's something that you never want to be reporting an incorrect value because your monitoring config is wrong. ES knows its cluster, and it seems like it should be the authority on it. Plus, I don't have a clean way that I would convey that through in my config (which gets generated from Consul service registrations).

What ES thinks it's name is isn't necessarily what it's name is within the rest of the system - targets should pretty much never chose what their labels are for monitoring. I have a setup where I've two distinct ES clusters with the same internal name for example.

discordianfish:
> What ES thinks it's name is isn't necessarily what it's name is within the rest of the system - targets should pretty much never chose what their labels are for monitoring. I have a setup where I've two distinct ES clusters with the same internal name for example.

Good point, we have the same actually.

elasticsearch_exporter.go
discordianfish: Given that one ES node (as in instance of ES running, not physical machine) can only be part of one cluster, I would prefer configuring it on the prometheus side as well. This is more explicit than having the exporter decide about those things on their own. I removed the cluster label for now.

I'm using consul as well and would use the relabling feature to set a cluster label from __meta_consul_service or __meta_consul_tags: http://prometheus.io/docs/operating/configuration/#consul-sd-configurations-consul_sd_config

@ewr Maybe that's a good/better option for you as well?

discordianfish:
So I'd be happy with this as it is. We can address 'rolling up' the timers / memory metrics later, same as having a global cluster view. Keep in mind this isn't working right now anyway. Current master also only exposes metrics for the local node.

elasticsearch_exporter.go
ewr: I'm fine with seconds. I had done ms to match what ES was reporting, but I don't have an attachment to it.

elasticsearch_exporter.go
```diff
@@ -267,55 +26,62 @@ type VecInfo struct {
...
-		"indices_docs_deleted":                      "Count of deleted documents on this node",
-		"indices_store_size_in_bytes":               "Size of stored index data in bytes",
-		"indices_segments_memory_in_bytes":          "Memory size of segments in bytes",
-		"jvm_mem_heap_committed_in_bytes":           "JVM heap memory committed",
-		"jvm_mem_heap_used_in_bytes":                "JVM heap memory used",
-		"jvm_mem_heap_max_in_bytes":                 "JVM heap memory max",
-		"jvm_mem_non_heap_committed_in_bytes":       "JVM non-heap memory committed",
-		"jvm_mem_non_heap_used_in_bytes":            "JVM non-heap memory used",
+		"indices_fielddata_memory_size_bytes":    "Field data cache memory usage in bytes",
+		"indices_filter_cache_memory_size_bytes": "Field data cache memory usage in bytes",
```
ewr: My error, but: this should be "Filter cache memory"

elasticsearch_exporter.go
```diff
@@ -267,55 +26,62 @@ type VecInfo struct {
...
+		"indices_docs":                           "Count of documents on this node",
+		"indices_docs_deleted":                   "Count of deleted documents on this node",
+		"indices_store_size_bytes":               "Current size of stored index data in bytes",
+		"indices_segments_memory_bytes":          "Current memory size of segments in bytes",
 	}
 	counterMetrics = map[string]string{
+		"indices_fielddata_evictions":           "Evictions from field data",
+		"indices_filter_cache_evictions":        "Evictions from field data",
```
ewr: Same thing here... Should be "from filter cache."

elasticsearch_exporter.go
ewr: Maybe this is just my lack of JVM understanding, but why should this metric be "seconds_count"? Just to match the metric with the seconds in it?

ewr:
I'm :+1: on all of this except the removal of node and cluster. I understand that they aren't needed in the happy case where you're running everything via best practices, but it complicates or removes two potential modes of operation: scraping the whole cluster and scraping a remote node.

Is the driver here just that we don't want extra labels being generated when they aren't necessary? Couldn't you overwrite or drop them via relabeling anyway?

brian-brazil:
> I understand that they aren't needed in the happy case where you're running everything via best practices, but it complicates or removes two potential modes of operation: scraping the whole cluster and scraping a remote node.

Scraping the cluster is a bit more complicated than adding some labels, you'll also need a special configuration on the Prometheus end to deal with the `instance` label clash.

If you must have the remote node different to the ES node, target discovery does allow controlling the `instance` separate from the address that's scraped.

> Couldn't you overwrite or drop them via relabeling anyway?

You could with a feature added earlier in the week, however that feature is intended only as a stopgap until you can fix the exporter.

ewr:
In the interest of pushing the ball forward on this, I've re-added node and cluster labels and merged everything else. I'm fine to continue talking about those, but I didn't want the other bits stuck in that conversation.

PR #7, opened 2015-08-20
fix collector vector sends

mattkanwisher:
Looks like your missing sending the counterVecs and gaugeVecs


ewr:
Thanks! I noticed that the other day when I needed them and they weren't there, but I hadn't been able to circle back to see where they went.

PR #13, opened 2016-02-15
Add cluster health metrics (fix #5)

### LINKED ISSUE #5: ES_exporter doesn't export cluster statistics
Currently only node level statistics are exported.  There's some valuable stats at /_cluster/health and /_cluster/stats that should be exported as well.

The key stats I noticed were missing:
- cluster status :exclamation: 
- index count
- total shard count
- primary shard count
- [relocating, initializing, unassigned] shard count

/_cluster/health:

```
{
    cluster_name: "devops-es",
    status: "green",
    timed_out: false,
    number_of_nodes: 7,
    number_of_data_nodes: 5,
    active_primary_shards: 22330,
    active_shards: 66990,
    relocating_shards: 0,
    initializing_shards: 0,
    unassigned_shards: 0
}
```

/_cluster/stats: http://pastebin.com/1nBbEH9T

FWIW this was my reasoning behind #1 - to reduce the amount of duplicated cluster-level statistics.  I'd like to implement a similar flag (--cluster-stats) to enable exporting these stats from only a subset of nodes in the cluster.

hudashot:
This introduces some metrics that only have 'cluster' label, so the whole
'gauges' vs 'gaugeVecs' distinction does not seem to make sense anymore.
I've defined all metrics and their labels explicitly.

### DIFF
--- a/elasticsearch_exporter.go
+++ b/elasticsearch_exporter.go
@@ -221,10 +221,6 @@ var (
 			help:   "Thread Pool current threads count",
 			labels: []string{"cluster", "node", "type"},
 		},
-		"cluster_status": &VecInfo{
-			help:   "Cluster status (0=green, 1=yellow, 2=red)",
-			labels: []string{"cluster"},
-		},
 		"cluster_nodes_total": &VecInfo{
 			help:   "Total number of nodes",
 			labels: []string{"cluster"},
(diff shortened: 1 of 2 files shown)


hudashot:
Happy to rebase this if you are interested to get it merged.

PR #21, opened 2016-07-27
Support ElasticSearch client certificate authentication

jrummler:
If ElasticSearch is protected by Shield/SearchGuard the exporter must do the client certificate authentication. This enables configuring this via PEM files containing the certificates / client cert / client private key.

Example call would be:

``` bash
elasticsearch_exporter --es.uri=elastic:9200 --es.ca=/path/to/pem/with/ca/cert.pem --es.client-private-key=/path/to/pem/with/client/private/key.pem --es.client-cert=/path/to/pem/with/client/cert.pem
```


deepthawtz:
👍  looks good. don't see much momentum on this project anymore which is unfortunate

metalmatze:
Hey. We've taken over the project and I would like to ask you to take a look at this PR once again.
If you're still happy with the code we can merge this after another review! 😊

jrummler:
Great news! I am still fine with it! Please get in touch if you need further support! ;-)

dominikschulz:
@jrummler Could you please rebase your PR? The changes itself look good, I'd like to merge them.

elasticsearch_exporter.go
```diff
@@ -429,3 +435,47 @@ func main() {
 	})
 	log.Fatal(http.ListenAndServe(*listenAddress, nil))
 }
+
+func createElasticSearchTlsConfig(pemFile, pemCertFile, pemPrivateKeyFile string) (*tls.Config) {
+	if (len(pemFile) > 0) {
```
metalmatze: This one can be inverted to `if (len(pemFile) <= ) { return nil }` so that we can get rid of the extra else at the bottom of this `func`. I think that applies for a few other conditions in this func too. It might be nitpicking, but definitely something I'd do different.

jrummler:
Guys, I have merged the changes. Since I do not use this tool anymore, I had no possibility to test the functionality.

PR #23, opened 2016-10-25
Put the protocol name in front of the hostname.

EdSchouten:
Without the protocol name this doesn't work. Adding this makes it
possible to start using the exporter purely looking at the
documentation.


EdSchouten:
I've just put in a workaround for #24 in this pull request as well.

PR #28, opened 2017-01-12
Add Dockerfile

dominikschulz:
This commit syncs the Makefile and Dockerfile with the ones currently
used in offical Prometheus exporters.


VERSION
```diff
@@ -0,0 +1 @@
+0.2.1
```
metalmatze: This one is not correct, right?

dominikschulz: This should reflect that latest release version. Isn't that 0.2.1?

dominikschulz: My bad, should have been 0.3.0 of course.

PR #30, opened 2017-02-06
Add file system and cluster-level statistics.

EdSchouten:
We'd like to keep track of whether Elasticsearch is failing due to full
disks. We already have these metrics declared in the structures. Let's
just export them as metrics as well. While there, also start scraping `/_cluster/health`, so we get access to global data.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -129,13 +129,63 @@ var (
 			labels: []string{"type"},
 		},
 	}
+
+	clusterHealthActivePrimaryShardsDesc = prometheus.NewDesc(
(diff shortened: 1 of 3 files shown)


metalmatze:
LGTM
I just wanted to give this a first test run. But I'm not against merging and testing before a release again.

metalmatze:
I just tested this and it's awesome. There are some really helpful metrics in here. Thank you very much!

EdSchouten:
And thank you for merging this. \o/

PR #38, opened 2017-04-03
Fix thread_pool_threads_count, thread_pool_largest_count, thread_pool_queue_count

wyukawa:
Now, thread_pool_active_count, thread_pool_threads_count, thread_pool_largest_count, thread_pool_queue_count are same.


wyukawa:
Thanks!

PR #44, opened 2017-05-05
Updates Dockerfile so that it can be independently used from make.

ghellings:
This is useful for creating an autobuild on Dockerhub. Running a make docker will still work as well.


dominikschulz:
This will include a full blown golang build environment in the image. The current build layout results in a much smaller package without unnecessary dependencies.

PR #53, opened 2017-05-25
Update Dockerfile

hantuzun:
Dockerfile was expecting a binary to be compiled on the host. However, such a binary could be dependent to the host environment. 

I updated the Dockerfile to create the binary in the image instead.


hantuzun:
We've published the latest image at Docker Hub: https://hub.docker.com/r/suggestgrid/elasticsearch_exporter/

dominikschulz:
We'd like to stay with the current build system. We'll probably move to multi-stage Dockerfiles in the future which solve this in a much cleaner way.

PR #58, opened 2017-06-08
Change time metrics from milliseconds to seconds

### LINKED ISSUE #57: Time metrics use Milliseconds instead of base unit - seconds
Great work on this exporter! we are using it with great results 👍 

One small comment, the following metrics use Milliseconds instead of a base unit (seconds)

```
elasticsearch_indices_flush_time_ms_total
elasticsearch_indices_indexing_index_time_ms_tot
elasticsearch_indices_merges_total_time_ms_total
elasticsearch_indices_refresh_total_time_ms_total
elasticsearch_indices_store_throttle_time_ms_total
elasticsearch_indices_search_query_time_ms
```

Is there any specific reason to use ms instead of sec?

I know I'm probably nit-picking here, but was reviewing the best practices from Prometheus and they seem to suggest to use base units, and leave handling the presentation in the Dashboarding tool (i.e. Grafana).

```
...should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers).
```
https://prometheus.io/docs/practices/naming/

Any thoughts?

dannyk81:
Should fix #57


metalmatze:
LGTM

PR #65, opened 2017-06-28
Split the exporter into multiple collectors

metalmatze:
We want to get rid of the big maps with counters and gauges.
This is a approach I could come up with.

Put everything a metric needs (type, desc, value, labels) into a struct to have it in one place. The improvement is, that these metrics, now have value and label `func`s to retrieve their values where they're declared. This way it's all combined at one place and not all over the place.

### Changes to metrics:
```diff
- elasticsearch_up
```
We're doing multiple requests concurrently to the endpoint now. Not really sure on how this metric could be helpful now. Would need some sort of global shared state between collectors. Dropping it for now.

```diff
- elasticsearch_cluster_health_status_is_green
- elasticsearch_cluster_health_status_is_red
- elasticsearch_cluster_health_status_is_yellow
```
These are duplicates of `elasticsearch_cluster_health_status` which has the color as a label, as it should be.

```diff
- elasticsearch_filesystem_data_free_percent
- elasticsearch_filesystem_data_used_percent
```
Percentage has been computed in this exporter up until now. We're dropping these metrics. This should be calculated in prometheus itself. We're probably going to provide recording rules that make this optional metrics.

```diff
- elasticsearch_indices_flush_time_seconds_total
+ elasticsearch_indices_flush_time_seconds
```
Because we're talking about seconds.

```diff
- indices_search_fetch_time_seconds_total
```
This is a duplicate of `indices_search_fetch_time_seconds`

```diff
- elasticsearch_indices_search_query_time_seconds_total
```
This is a duplicate of `indices_search_query_time_seconds`


collector/cluster_health.go
```diff
@@ -0,0 +1,246 @@
...
+			[]string{"cluster"}, nil,
+		),
+		NumberOfNodes: prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, subsystem, "number_of_nodes"),
+			"Number of nodes in the cluster.",
+			[]string{"cluster"}, nil,
+		),
+		NumberOfPendingTasks: prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, subsystem, "number_of_pending_tasks"),
+			"XXX WHAT DOES THIS MEAN?",
```
dominikschulz: Intended or left over? If we can't figure out a proper description we may want to put an `undocumented` instead.

collector/cluster_health.go
```diff
@@ -0,0 +1,246 @@
...
+			[]string{"cluster"}, nil,
+		),
+		ActiveShards: prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, subsystem, "active_shards"),
+			"Aggregate total of all shards across all indices, which includes replica shards.",
+			[]string{"cluster"}, nil,
+		),
+		DelayedUnassignedShards: prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, subsystem, "delayed_unassigned_shards"),
+			"XXX WHAT DOES THIS MEAN?",
```
dominikschulz: document, please

collector/cluster_health.go
```diff
@@ -0,0 +1,246 @@
...
+			[]string{"cluster"}, nil,
+		),
+		StatusIsRed: prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, subsystem, "status_is_red"),
+			"Whether all primary and replica shards are allocated.",
+			[]string{"cluster"}, nil,
+		),
+		TimedOut: prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, subsystem, "timed_out"),
+			"XXX WHAT DOES THIS MEAN?",
```
dominikschulz: document, please

collector/cluster_health.go
```diff
@@ -0,0 +1,246 @@
...
+
+	ch <- prometheus.MustNewConstMetric(
+		c.RelocatingShards,
+		prometheus.GaugeValue,
+		float64(clusterHealthResponse.RelocatingShards),
+		cluster,
+	)
+
+	var statusValue float64
+	if clusterHealthResponse.Status == "green" {
```
dominikschulz: What about red and yellow? Would like to be able to distinquish between those two as well.

collector/cluster_health.go
```diff
@@ -0,0 +1,246 @@
...
+			prometheus.BuildFQName(namespace, subsystem, "number_of_pending_tasks"),
+			"XXX WHAT DOES THIS MEAN?",
+			[]string{"cluster"}, nil,
+		),
+		RelocatingShards: prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, subsystem, "relocating_shards"),
+			"The number of shards that are currently moving from one node to another node.",
+			[]string{"cluster"}, nil,
+		),
+		StatusIsGreen: prometheus.NewDesc(
```
dominikschulz: Could we drop this, please?
We already have the status metric below.

collector/cluster_health.go
```diff
@@ -0,0 +1,246 @@
...
+	ActivePrimaryShards     *prometheus.Desc
+	ActiveShards            *prometheus.Desc
+	DelayedUnassignedShards *prometheus.Desc
+	InitializingShards      *prometheus.Desc
+	NumberOfDataNodes       *prometheus.Desc
+	NumberOfInFlightFetch   *prometheus.Desc
+	NumberOfNodes           *prometheus.Desc
+	NumberOfPendingTasks    *prometheus.Desc
+	RelocatingShards        *prometheus.Desc
+	StatusIsGreen           *prometheus.Desc
```
dominikschulz: Can we drop this in favor of Status?

collector/cluster_health.go
```diff
@@ -0,0 +1,246 @@
...
+	Status                  *prometheus.Desc
+	StatusIsYellow          *prometheus.Desc
```
dominikschulz: Can we drop this in favor of Status?

collector/cluster_health.go
```diff
@@ -0,0 +1,246 @@
...
+	StatusIsRed             *prometheus.Desc
```
dominikschulz: Can we drop this in favor of Status?

collector/cluster_health.go
metalmatze: I currently assume that `elasticsearch_cluster_health_status` is `0` if red or yellow and only `1` when green.
I can only think about doing something like 
* `red` => `0`
* `yellow` => `1`
* `green` => `2` 
which honestly feels weird.

dominikschulz: Yes, indeed. I see that issue. But on the other hand I'd like to differentiate between red and yellow ...

PR #70, opened 2017-06-30
Fixed docker-compose example

algestam:
I didn't get the exporter to find the elasticsearch cluster when following the docker-compose example at first but then I noticed that the command argument was declared as an environment variable. Changing that fixed it for me.


metalmatze:
Whoops! Thanks for that fix!

PR #73, opened 2017-07-06
Ship metrics about scrape timestamps and counts.

jmcarp:
Allow users to alert when the scraper either hasn't run for a while or isn't able to query the cluster health or nodes endpoints. This is largely taken from https://github.com/cloudfoundry-community/bosh_exporter by @frodenas.

Pairing with @rogeruiz

### DIFF
--- a/collector/nodes.go
+++ b/collector/nodes.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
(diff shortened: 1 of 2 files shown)


collector/nodes.go
```diff
@@ -991,13 +1054,18 @@ func (c *Nodes) fetchAndDecodeNodeStats() (nodeStatsResponse, error) {
 }
 
 func (c *Nodes) Collect(ch chan<- prometheus.Metric) {
+	begun := time.Now()
```
metalmatze: Why not call it `start := time.Now()`?

metalmatze:
In general I think that these metrics show follow the style we started in this version. So don't introduce them all on their own but add them to the slice of metrics we'll iterate over later on.
This just makes it way more maintainable.

I could look into this one on Monday and make proper improvements to this.

jmcarp:
I'd be happy to revise this, but I'm not sure how to apply the pattern of iterating over a slice of metrics whose values are all a function of the same input here. Unlike the slice of cluster health metrics, which all do a calculation on the cluster health response, the new metrics don't all operate on the same data.

dominikschulz:
You both have a point there. These metrics are different than the existing ones. But still I'm not yet happy with this PR.

I need some more time to think about this, but I'm sure we'll find a suitable approach next week.

jmcarp:
@dominikschulz @metalmatze: I'd be happy to revise this--let me know if you prefer a different a different style for the new metrics.

PR #78, opened 2017-07-14
Expose scraping metrics

### LINKED ISSUE #48: Measure ElasticSearch API timeouts / Check elasticsearch_up
In some cases the `elasticsearch_up` metric didn't work as expected.

We need to make sure this essential metric is working as expected.

dominikschulz:
This PR adds metrics about the scrapes themselfes in a way similar to the (official) HAProxy exporter:
https://github.com/prometheus/haproxy_exporter/blob/master/haproxy_exporter.go.

I'm not quite happy with this PR, but I think this may be the best way to do it. Would be happy to be shown a better way. IMHO it's a little cleaner than #73 

Fixes #48

### DIFF
--- a/collector/cluster_health.go
+++ b/collector/cluster_health.go
@@ -38,6 +38,9 @@ type ClusterHealth struct {
 	client *http.Client
 	url    *url.URL
 
+	up                              prometheus.Gauge
+	totalScrapes, jsonParseFailures prometheus.Counter
+
 	metrics      []*clusterHealthMetric
 	statusMetric *clusterHealthStatusMetric
 }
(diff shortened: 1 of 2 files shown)


collector/cluster_health.go
```diff
@@ -217,21 +237,31 @@ func (c *ClusterHealth) fetchAndDecodeClusterHealth() (clusterHealthResponse, er
...
+	c.totalScrapes.Inc()
+	defer func() {
+		ch <- c.up
+		ch <- c.totalScrapes
+		ch <- c.jsonParseFailures
+	}()
+
 	clusterHealthResponse, err := c.fetchAndDecodeClusterHealth()
 	if err != nil {
+		c.up.Set(0)
```
zwopir: this is not completely correct - the cluster endpoint might answer, but json encoding fails

collector/nodes.go
```diff
@@ -980,26 +999,35 @@ func (c *Nodes) fetchAndDecodeNodeStats() (nodeStatsResponse, error) {
...
+	c.totalScrapes.Inc()
+	defer func() {
+		ch <- c.up
+		ch <- c.totalScrapes
+		ch <- c.jsonParseFailures
+	}()
+
 	nodeStatsResponse, err := c.fetchAndDecodeNodeStats()
 	if err != nil {
+		c.up.Set(0)
```
zwopir: dito

collector/cluster_health.go
dominikschulz: For certain definitions of correct ;)
It really depends how you define `up` in this case. But I'm completely open to suggestions. How should we define `up`? I used it as `the cluster is reachable via HTTP and answering with an HTTP Status 200`. Seems to be the same definition used by the HAProxy exporter. For the case that JSON decoding fails we'd still have the other metric.

But we can change the definition of up, of course. As longs as it's more or less consistent with other exporters.

dominikschulz: Oh, I see ... this slightly changed after rebasing onto the current master.

What's your suggestion here?

PR #85, opened 2017-08-16
add index metrics

matsumana:
I would like to monitor index metrics, just like Kinaba X-pack monitoring.
so I added 3 metrics below:

```
# HELP elasticsearch_indices_docs_primary Count of documents which only primary shards
# TYPE elasticsearch_indices_docs_primary gauge
elasticsearch_indices_docs_primary{cluster="test",index="foo_1"} 2
elasticsearch_indices_docs_primary{cluster="test",index="foo_2"} 3

# HELP elasticsearch_indices_store_size_bytes_primary Current total size of stored index data in bytes which only primary shards on all nodes
# TYPE elasticsearch_indices_store_size_bytes_primary gauge
elasticsearch_indices_store_size_bytes_primary{cluster="test",index="foo_1"} 8425
elasticsearch_indices_store_size_bytes_primary{cluster="test",index="foo_2"} 12420

# HELP elasticsearch_indices_store_size_bytes_total Current total size of stored index data in bytes which all shards on all nodes
# TYPE elasticsearch_indices_store_size_bytes_total gauge
elasticsearch_indices_store_size_bytes_total{cluster="test",index="foo_1"} 16850
elasticsearch_indices_store_size_bytes_total{cluster="test",index="foo_2"} 24840
```

PR #86, opened 2017-08-16
add index metrics via alias

matsumana:
__Please review and merge #85 first.__

<br>

I would like to monitoring whether alias settings are correct.
I realize it to monitoring index metrics via alias.

so I added the metrics below:

```
# HELP elasticsearch_aliases_docs_primary Count of documents which only primary shards via alias
# TYPE elasticsearch_aliases_docs_primary gauge
elasticsearch_aliases_docs_primary{alias="foo",cluster="test",index="foo_1"} 2
elasticsearch_aliases_docs_primary{alias="fooA",cluster="test",index="foo_1"} 2
elasticsearch_aliases_docs_primary{alias="fooB",cluster="test",index="foo_2"} 3
```


matsumana:
Hello, Please also review this PR.

collector/indices_test.go
```diff
@@ -0,0 +1,54 @@
...
...8691,"total_docs":23973218,"total_size_in_bytes":19660778820,"total_stopped_time_in_millis":0,"total_throttled_time_in_millis":0,"total_auto_throttle_in_bytes":41943040},"refresh":{"total":34871,"total_time_in_millis":321333,"listeners":0},"flush":{"total":0,"total_time_in_millis":0},"warmer":{"current":0,"total":34873,"total_time_in_millis":7932},"query_cache":{"memory_size_in_bytes":0,"total_count":17,"hit_count":0,"miss_count":17,"cache_size":0,"cache_count":0,"evictions":0},"fielddata":{"memory_size_in_bytes":0,"evictions":0},"completion":{"size_in_bytes":0},"segments":{"count":20,"memory_in_bytes":429521,"terms_memory_in_bytes":251138,"stored_fields_memory_in_bytes":17536,"term_vectors_memory_in_bytes":0,"norms_memory_in_bytes":1280,"points_memory_in_bytes":37815,"doc_values_memory_in_bytes":121752,"index_writer_memory_in_bytes":0,"version_map_memory_in_bytes":0,"fixed_bit_set_memory_in_bytes":0,"max_unsafe_auto_id_timestamp":-1,"file_sizes":{}},"translog":{"operations":249730,"size_in_bytes":310931832},"request_cache":{"memory_size_in_bytes":0,"evictions":0,"hit_count":0,"miss_count":1},"recovery":{"current_as_source":0,"current_as_target":0,"throttle_time_in_millis":0}}}}}`,
```
zwopir: can you please add fixtures for 1.7.6 and 2.4.5 (compare nodes_test.go)?

matsumana:
@zwopir 
I added fixtures for 1.7.6, 2.4.5 and 5.4.2
Would you please review again?

zwopir:
thanks for the update. I'm not really happy with the PR:
 - the collector does two API calls, one of them is redundant - at least if you restructure the code. I haven't yet thought how to refactor
 - within the alias collector you use results from the `_alias`API call as map key for the results of the `_all/_stats` call. This should be consistent ES wise, but I wouldn't rely on it.

Could you elaborate a bit more on the use case of this PR? I would like to keep the exporter and especially the collectors as well structured and uncoupled as possible

matsumana:
For example, There are the following index and alias.

- foo_1 (index)
- foo (alias)

An application refers to the index via alias.
However there is a possibility that it can not be referred due to a program bug or misoperation.
so I want to monitoring whether alias settings are correct.

I think I can realize it, by this metric(elasticsearch_aliases_docs_primary) and following PromQL.

```
absent(elasticsearch_aliases_docs_primary{alias="foo"}) == 1
```

zwopir:
I understand your use case, but I don't think its a use case that should be monitored via metrics. Misconfiguration is generally speaking not a thing you want to monitor constantly (i.e. with a scape interval in the range of seconds), but just after a change within the ES cluster or after a software release that changes the index/alias settings. Especially if it is done with a "nagios style" ok/not ok metric. So it's more a matter of release management and QA, not of operations. 
I would therefore reject this PR.

What do you think, @dominikschulz and @falschparker82 (speaking a bit more from the ES side)?

matsumana:
I understand and agree with you.
I will consider other way.

PR #93, opened 2017-09-19
Fix container version

yanns:


yanns:
He Christoph! The world is small. 😉 
I'll take a look at the new version, thx for the hint.
This exporter is very useful 👍

zwopir:
nice to hear :) and yes, the world is small - at least the IT world...

PR #112, opened 2017-12-07
Fix formatting in logs.

tgxworld:


collector/cluster_health.go
```diff
@@ -227,7 +227,7 @@ func (c *ClusterHealth) fetchAndDecodeClusterHealth() (clusterHealthResponse, er
 	u.Path = "/_cluster/health"
 	res, err := c.client.Get(u.String())
 	if err != nil {
-		return chr, fmt.Errorf("failed to get cluster health from %s://%s:%s/%s: %s",
+		return chr, fmt.Errorf("failed to get cluster health from %s://%s:%s%s: %s",
```
dominikschulz: `u.Path` already incldudes a leading `/`?!

tgxworld: Yup see https://github.com/justwatchcom/elasticsearch_exporter/pull/112/files#diff-a814f28c8fbafb929efda90ac8ce519dL227

PR #113, opened 2017-12-08
Expose load averages

pchristos:
Tested against versions 5.1.2 - 5.6.4

### DIFF
--- a/collector/nodes.go
+++ b/collector/nodes.go
@@ -150,6 +150,42 @@ func NewNodes(logger log.Logger, client *http.Client, url *url.URL, all bool) *N
 			{
 				Type: prometheus.GaugeValue,
 				Desc: prometheus.NewDesc(
+					prometheus.BuildFQName(namespace, "os", "load1"),
+					"Shortterm load average",
+					defaultNodeLabels, nil,
+				),
+				Value: func(node NodeStatsNodeResponse) float64 {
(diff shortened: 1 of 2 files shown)


dominikschulz:
I'm not sure if this should be in this exporter. It looks more like something you'd want to do with the node exporter.

pchristos:
What if someone has his Elasticsearch hosted aaS with no access to the nodes? Node exporter would have to be deployed on the nodes themselves, right? Plus, metrics like CPU, which are not Elasticsearch-specific, are already being exported.

zwopir:
@dominikschulz is right about that the data itself is redundant with the metrics provided by the node exporter. But the data is exported by the nodes endpoint anyways, so it doesn't make a big difference to create a metric from it or not. On the plus side:
 - you get at least some OS metrics on hosted ES
 - the OS metrics exported via the elasticsearch_exporter have the same labels as the ES metrics

so I'd vote for accepting this PR.

@pchristos could you add some tests in `nodes_test.go` that you can access the necessary fields in all supported versions?! we have fixture JSONs inline in the test file, so it's just a matter of checking that the struct fields are not nil or the type initialization value

dominikschulz:
SGTM

zwopir:
just merged. Thanks for your contribution, @pchristos !

PR #118, opened 2018-01-02
Fix descriptions of indexing metrics

janko-m:
Every other `*_seconds_total` metric measures the time it took to do some action, so I was first surprised that with `elasticsearch_indices_indexing_index_*` metrics it was the other way around, but then the code confirmed that it was just a typo in the README.


zwopir:
Hi @janko-m,

thanks for the PR!

PR #123, opened 2018-01-28
Add support for shard-level document count

mtritschler:
In cases where replicas get out of sync with the primary, it would be extremely useful to be able to track differences in the count of document on shard-level over time.

### DIFF
--- a/collector/indices.go
+++ b/collector/indices.go
@@ -25,6 +25,15 @@ type indexMetric struct {
 	Labels func(indexName string) []string
 }
 
+type shardMetric struct {
+	Opts   prometheus.GaugeOpts
+	Type   prometheus.ValueType
+	Desc   *prometheus.Desc
+	Value  func(data IndexStatsIndexShardsDetailResponse) float64
+	Labels []string
+	LabelValues func(indexName string, shardName string, data IndexStatsIndexShardsDetailResponse) prometheus.Labels
+}
+
 type Indices struct {
 	logger log.Logger
 	client *http.Client
@@ -35,6 +44,7 @@ type Indices struct {
 	jsonParseFailures prometheus.Counter
 
 	indexMetrics []*indexMetric
+	shardMetrics []*shardMetric
 }
 
 func NewIndices(logger log.Logger, client *http.Client, url *url.URL) *Indices {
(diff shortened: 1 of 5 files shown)


zwopir:
Hi @mtritschler ,

thanks for your PR. I had a quick view over your PR and code-wise it looks good so far. Unfortunately I don't know what the ` level=shards` parameter does to the returned output and how it differs from the output retrieved without the parameter. Can you provide us with the output of the curl query with and without the parameter? Feel free to add them as additional testdata in our testcases

mtritschler:
I put both types of responses in [this gist](https://gist.github.com/mtritschler/0641527c89195795d151026547a1526d) (grabbed from a dockerized ES 5.6.6). Note that in an actual cluster, there will be more than one element in the list found at `indices.<index_name>.shards.<shard_number>`.

There are some assumptions built into your test data (at least regarding the number of docs AFAICS), which are not reflected in this setup, so I could not just replace the existing test data with this output. Let me know how to reproduce and I will update the PR @zwopir

mtritschler:
Thinking more about this (e.g. how tracking deletes per shard would change things), `elastic_indices_docs_shard` should probably become `elastic_indices_shard_docs`. This would make it clear that everything starting with `elastic_indices_shard_*` is a shard-level metric. I will update the PR accordingly

zwopir:
yes, that makes sense. Since the json output may get quite big in a large cluster, it probably makes sense to provide a toggle that enables or disables the shard metrics. Would you mind adding such a toggle?!

mtritschler:
I added `es.shards` using `es.nodes` as blueprint. For the sake of brevity I let `es.shards=true` imply `es.indices=true`, because otherwise setting just `es.shards=true` would have had no noticeable effect on its own.

zwopir:
Hi @mtritschler ,

sorry it took so long. Your PR looks good to merge, I'll just ask my colleague to have a quick look. I planned to build a rc for later this week. It will in include your code

zwopir:
Hi @mtritschler ,

could you please run a go fmt on main.go?!

mtritschler:
@zwopir thanks for reviewing. I have fixed the issue with formatting

mtritschler:
Any update on when this might get merged into master?

collector/indices.go
```diff
@@ -423,6 +472,18 @@ func (i *Indices) Collect(ch chan<- prometheus.Metric) {
 				metric.Value(indexStats),
 				metric.Labels(indexName)...,
 			)
+
+		}
+		if i.shards {
```
dominikschulz: I'd write this as `if !i.shards { continue }` so you could outdent the following loop, but it's OK to leave it as is.

PR #125, opened 2018-02-16
Add node-level current merge metrics

schallert:
Hey folks, I was using this metric to debug an issue recently and thought it might be helpful for others. This adds a gauge to track current number of merges on a per-node basis.


zwopir:
Hi Matt, thanks for you contribution, looks good!

PR #129, opened 2018-02-23
Add OS CPU percent metric

dashford:
We're using an elastic.co hosted cluster and noticed that the `cpu_percent` metric was reporting as `0` consistently (coming from the `process` object. This PR adds the `cpu_percent` metric for the `OS` object.

I've also added the existing `elasticsearch_os_load1`, `elasticsearch_os_load5`, and `elasticsearch_os_load15` metrics to the `README`.

Let me know if I'm on the right track with this or if tests need to accompany it.


zwopir:
Hi @dashford 

thanks for the contribution. I try to create a release later this week that includes your code

dashford:
Thanks @zwopir!

PR #138, opened 2018-04-20
Use `sum/count` instead of just `sum` to catch elasticsearch_cluster_health_number_of_nodes

andreyev:
If you have a exporter in each ES node and just `sum` all `elasticsearch_cluster_health_number_of_nodes` you will get a square of your cluster size... :grimacing:


zwopir:
thanks for the catch and the fix!

PR #141, opened 2018-04-23
Fixes #140

### LINKED ISSUE #140: Injecting Secrets for the es.uri
The es.uri requires basic auth for the username and password.

Within kubenetes, secrets can only be injected using a file or a environment variable. The current elasticsearch_exporter docker container doesn't allow the es.uri to be injected using an environment variable which means that the secret for the basic auth can't be injected.

I need an environment variable (something like `ES_URI`) which could be read by the docker container and supplies the es.uri flag with the appropriate settings.

lswith:


zwopir:
thanks for your contribution!
if the variable is only needed, inside the if branch, I prefer
```
if esURIEnv, ok := os.LookupEnv("ES_URI"); ok {
...
```
but it's completely fine this way. I might change the cmdline parsing to kingpin anyways which includes environment variable bindings by default, so there's no need to change it

Issue #135, opened 2018-04-04
elasticsearch_indices_docs_primary metric is missing labels (cluster, es_***_node, etc..)


dannyk81:
I'm using v1.0.2 of the exporter and looks like `elasticsearch_indices_docs_primary` is missing all the "default" labels that we find in other metrics: `cluster`, `es_***_node`, `host`

dannyk81:
ping @zwopir 😄

zwopir:
actually, it's not really a bug. The indices metrics are querying
`http://<host>:9200/_all/_stats?level=shards`.
This endpoint doesn't contain the values for the "default" labels, only the
node_id (some ES internal entity without external meaning). We've
implemented the metric classes as independent prometheus collectors. The
each call different ES API endpoints, each providing a different set of
information, so I don't really see a convenient way to correlate these
collectors without completely rewriting the exporter.

Can you elaborate on your use case so we can find a solution?!

dannyk81:
We have multiple Elasticsearch clusters scraped by the same Prometheus server, and we use the `cluster` label in our queries to isolate the metrics per cluster.

Not having at least the `cluster` label makes it difficult to easily filter these metrics.

Currently I'm using a PromQL "hack" where I group these metrics with another (node) metric based on the instance label and extract the cluster label from there, it seems to work but was hoping to have these labels natively from the exporter.

zwopir:
ok, I understand. It makes sense. To be honest, setting up a elasticsearch_exporter for each cluster seems like the easiest solution. The collectors that have/need label values from each other are completely independent. So correlating the label values would require some serious refactoring that doesn't make sense from any other point of view.

Maybe excluding the promQL-hack to recorded queries simplifies that usage of these metrics...?!

dannyk81:
We are running elasticsearch_exporter on every elastic node, but they are all (nodes from all clusters) defined under the same job because we use Puppet based service discovery to build the targets list.

I already created a recording rule with the query I mentioned and we started using that in our dashboards, so not a big deal, though if this could be somehow fixed on the exporter level it would be stellar.

zwopir:
ok, I understand your use case for the labels. I can't promise a short-term fix for your issue though, but I'll keep it in mind for a next refactoring round.

zwopir:
Hi @dannyk81,

[...]

Issue #131, opened 2018-03-12
Count of indices


j0hnsmith:
There doesn't seem to be a 'total # of indices in cluster' metric, this would be useful.

zwopir:
Hi @j0hnsmith,

I've checked the ES API endpoints we are currently querying on each prometheus scrape. They don't return the requested metric. So I would have to query another endpoint to get the metric. It would be something like `http://localhost:9200/_cat/indices?format=json&pretty`. I think the number of indices is a metric that doesn't change frequently. So adding another http call on each scrape for a more or less constant metric doesn't seem appropriate to me

j0hnsmith:
Fair point, does depend on usage though. Is there any scope to make it optional, eg off by default but users can enable?

zwopir:
yes, that'd be certainly possible. It's quite a quite easy task, the json returned by the endpoint pretty clear. See https://www.elastic.co/guide/en/elasticsearch/reference/current/cat.html for additional information how to change the units of the returned metrics (bytes vs. human, ...) So if we (you?) implement it we just need to make sure the queries work on all supported versions, i.e. 1.8.x, 2.x, 5.x and 6.x. and that the json looks the same. If it doesn't, I think it's not worth the effort.

So I can review a PR that optionally adds the feature and/or help with it, but working on it myself hasn't got the highest priority

benkeil:
If you use daily indices its not a constant size.

gavk34:
I might have time to look into this - seems the `_all/_stats` has this information no?
Should be just a matter of capturing the root(name of index) to a metric - which users can sum?
Example:
curl -s https://my-elasticsearch-url.com/_all/_stats\?format\=json\&pretty | jq '.indices'
```
"kubernetes-2018.08.27": {
    "primaries": {
    ...
```

My Go is not expert level so I wouldn't be expecting this urgently however we have use ourselves for such a metric

j0hnsmith:
@gavk34 see the PR and further discussion in #143.

zwopir:
thanks for the clarification, @j0hnsmith 
closing this issue

TL;DR; 
count of indices can be obtained with `count by (instance) (elasticsearch_indices_docs_primary)`

PR #143, opened 2018-04-25
add count of indices in cluster

j0hnsmith:
Realised that `/_all/_stats` contains _all_ indices in the cluster, ref #131.

### DIFF
--- a/main.go
+++ b/main.go
@@ -83,7 +83,9 @@ func main() {
 			os.Exit(1)
 		}
 		defer resp.Body.Close()
-		d := new(bodyData)
+		d := &struct {
+			ClusterName string `json:"cluster_name"`
+		}{}
 		if err := json.NewDecoder(resp.Body).Decode(d); err != nil {
 			level.Error(logger).Log(
 				"msg", "error decoding cluster data",
@@ -108,10 +110,6 @@ func main() {
 			"err", err,
 		)
 	}
-}
-
-type bodyData struct {
-	ClusterName string `json:"cluster_name"`
 }
 
 // IndexHandler returns a http handler with the correct metricsPath
--- a/main.go
+++ b/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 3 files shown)


dominikschulz:
Thanks for the pull request!

IMHO you shouldn't pull in the cluster name. It can change during the runtime of the exporter if e.g. a different node/configuration is made available at the same destination address.

j0hnsmith:
Yeah, I thought about that too. Docs say to change the cluster name you should perform a full cluster restart, I'm _assuming_ (quite a big assumption) that will include the exporter. Changing the cluster name should be a very rare event, it seemed better than an extra http request on every scrape. What do you think? Any alternative ideas?

zwopir:
Hi @j0hnsmith,

I just talked to Dominik and agree with him regarding his comment. For now my idea would be to just not include the cluster name as a constant label. I definitely don't like the idea getting the cluster name on every scrape. On the other hand outdated cluster name information in the labels are more harmful than not having them at all.

There's an issue (#135) also addressing the "cluster name as label" topic. So re-adding the label can be done while working on #135.

Maybe a label updater go routine that regularly gets the labels might be an option to solve the problem. Seems a bit to much logic involved for a simple exporter, thought (i.e. getting the updates in the collectors). What do you think, @dominikschulz ?

j0hnsmith:
Many people run multiple ElasticSearch clusters, in this case metrics without a cluster label will be plain wrong. Feels like it's better to resolve #135 before moving forward with this.

dominikschulz:
IMHO the exported metrics should not include the cluster name at all.
You can and should add that during relabeling.

j0hnsmith:
Relabeling requires additional config, I don't understand how/why that improves the situation, can you elaborate?

dannyk81:
Quite agree with @j0hnsmith here, including this information during relabeling forces us to have some way to sync the Cluster name we define in Elasticsearch and Prometheus configuration, this is very tricky and having this identifier (i.e. Cluster name) provided from the application via the exporter is akin to providing the index names, I do hope you will reconsider your stance.

main.go
```diff
@@ -92,6 +110,10 @@ func main() {
 	}
 }
 
+type bodyData struct {
```
metalmatze: You can put this type into the func itself right before it is used. It seems to be the only place where you use it.

Something like:
```go
bodyData := &struct {
	ClusterName string `json:"cluster_name"`
}{}
```

maxromanovsky:
Nice PR! How about adding also the health of index? Maybe as simple as green/red/yellow?

caarlos0:
count of indices can be obtained with `count by (instance) (elasticsearch_indices_docs_primary)`

gavk34:
@j0hnsmith Can we possibly add the index name as a label to a metric, that way users can simply use Prometheus to sum up that value to give them the count of indices and also the index name - granted we are running a fairly lightweight cluster so not too sure of the cardinality of this.

j0hnsmith:
@gavk34 sounds like you want per index metrics eg name, size, doc count, I think you're looking for the `es.indices` flag.

gavk34:
Ah brilliant @j0hnsmith, we have `es.all` which I falsely believed did everything 🤦‍♂️ 

My bad I should have read the documentation - sorry for the noise 👍

zwopir:
obsoleted by #176

PR #153, opened 2018-06-02
feat: added more index metrics

### LINKED ISSUE #152: per-index metrics
Hi, thanks for the great work on this exporter!

So, it seems that the exporter query `/_all/_stats`, which returns all stats indeed 😄 

Those stats include:

- number of documents of all indices
- number of deleted documents of all indices
- index size in bytes

and others, but these 3 are the ones I miss :P 

Would you be interested in a PR adding support for reporting those metrics as well?

They may be interesting, for example, if the `rate(elasticsearch_index_deleted_document_count[5m]) by (index_name) > 1234` that means that  something is deleting a lot of documents or whatever...

Index size may also be useful to have the history of the index size and decide if we should reindex it with more or less shards and etc..

Anyway, let me know what you think. If you're interested, I'll work on a PR.

Cheers!

caarlos0:
closes  #152

### DIFF
--- a/collector/indices.go
+++ b/collector/indices.go
@@ -84,6 +84,42 @@ func NewIndices(logger log.Logger, client *http.Client, url *url.URL, shards boo
 			{
 				Type: prometheus.GaugeValue,
 				Desc: prometheus.NewDesc(
+					prometheus.BuildFQName(namespace, "indices", "deleted_docs_primary"),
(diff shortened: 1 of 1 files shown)


zwopir:
Thanks for your contribution, Carlos!

caarlos0:
thanks =D

any ETA for a new release?

PR #155, opened 2018-06-06
Add elastic max file descriptors metric

joaosilva15:
This metric would be useful to have when looking at elastic search file descriptors.


zwopir:
LGTM, thanks!

PR #159, opened 2018-07-13
Support for reverse proxies

jmdacruz:
Currently, both `cluster_health.go` and `nodes.go` are overwriting the path provided by the `es.uri` flag, assuming that it will always be empty. However, this assumption is not always correct: In some cases it's useful to access the Elasticsearch endpoint behind a reverse proxy (e.g., Kubernetes) for testing purposes. This fix adds support for this scenario.

### DIFF
--- a/collector/cluster_health.go
+++ b/collector/cluster_health.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"path"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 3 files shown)


jmdacruz:
Just realized I missed `indices.go`. Working on the fix

zwopir:
Hi,

thanks for your contribution! It makes sense to support reverse proxying. I'm currently on holidays with my kids, so I can't review it before the weekend

zwopir:
looks good, thanks for your contribution!

PR #160, opened 2018-07-16
Add option to query a specific node from cluster

joaosilva15:
With the current options of the elastic exporter we had trouble gather metrics for some nodes of the cluster that hadn't the http api enabled.  This made gathering metrics for this nodes only possible by activating the option to gather all the cluster metrics. This, for us, was not the desired option.

To solve this problem we used the option of the elastic api to pass as an argument the name of the node and receive the metrics for that node.

With this change we can query our nodes with the exporter and receive the metrics for the nodes of the cluster that don't have the http api installed.

### DIFF
--- a/collector/nodes.go
+++ b/collector/nodes.go
@@ -128,6 +128,7 @@ type Nodes struct {
 	client *http.Client
 	url    *url.URL
 	all    bool
+	node   string
 
 	up                              prometheus.Gauge
 	totalScrapes, jsonParseFailures prometheus.Counter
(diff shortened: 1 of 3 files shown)


zwopir:
Hi,

thanks for your contribution! It makes sense. I'm currently on holidays with my kids, so I can't review it before the weekend

joaosilva15:
Rebased and es.all flag documentation improved.

zwopir:
thanks for the contribution, Joao!

PR #162, opened 2018-07-17
Include cluster name in -es.indices or -es.shards output.

### LINKED ISSUE #135: elasticsearch_indices_docs_primary metric is missing labels (cluster, es_***_node, etc..)

astral303:
Cluster name is set once the cluster can be connected to, and is not
refreshed after that. Fixes #135 for my own purposes by adding the most important label: `cluster`.

### DIFF
--- a/collector/cluster_health.go
+++ b/collector/cluster_health.go
@@ -220,6 +220,20 @@ func (c *ClusterHealth) Describe(ch chan<- *prometheus.Desc) {
 	ch <- c.jsonParseFailures.Desc()
 }
 
+func (c *ClusterHealth) FetchClusterName() (string, error) {
+	clusterHealthResponse, err := c.fetchAndDecodeClusterHealth()
+	if err != nil {
+		c.up.Set(0)
+		level.Warn(c.logger).Log(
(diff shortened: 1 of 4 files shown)


zwopir:
Hi @astral303,

thanks for the PR! I'm somehow not really convinced how you solved this problem. I already proposed a solution in !135 but haven't got the time to work on it. My arguments are
- the cluster health endpoint is queried twice per scrape for a value that rarely changes.
- If we need to call a second endpoint in the indices collector for the cluster name, I'd chose the `/` ES route, not the cluster health route.

What do you think about adding a pointer to channel to the `ClusterHealth` collector struct and passing an externally created channel. Each time its method `fetchAndDecodeClusterHealth` is called, we can write the cluster name (and other relevant) labels to that channel. The `fetchClusterName func() string` in the `NewIndices` constructor can be changed to read from that channel.

zwopir:
no feedback on my request. Implemented it myself in #176

PR #163, opened 2018-07-17
Add many more per-index stats, such as query time metrics.

astral303:
This add many important per-index metrics to the `-es.indices` output, including per-index operation timing and throughput.

Due to conflicts with the non-indices version of the metrics, I ended up having to use `index_stats` instead of `indices` in the metric names.


zwopir:
thanks for you contribution! merged

PR #164, opened 2018-07-19
Linter fixes

mjtrangoni:
Hi @zwopir , @dominikschulz,

I fixed some trivial linter issues, that I found with [gometalinter](https://github.com/alecthomas/gometalinter) and [codespell](https://github.com/codespell-project/codespell). 
There is still more issues, but I want to ask you first if you are interested on getting those fixes at all.

BTW, I think you can implement the [prometheus/common log and version](https://github.com/prometheus/common) libraries, as well as [kingpin](https://github.com/alecthomas/kingpin) instead of golang's flag.

Another big approach would be using [dep](https://github.com/golang/dep) instead of `govendor`.

Just let me know if you are interested in some of them.


mjtrangoni:
Ping @zwopir

zwopir:
Hi @mjtrangoni,

thanks for your PR! I would appreciate it if you fix the other linter issues as well without putting too much priority into it.
Regarding kingpin: I'm fine with the stdlib flag at the moment. Introducing kingpin would change the single dash `-` arguments to double dash`--` arguments. 'm planning a breaking change release (changing the port) after the recent changes are stable and releases as non-RC release.
Regarding logging: I prefer go-kit over the prometheus log package (which is a wrapper around logrus)

I didn't quite get what you mean with prometheus/common/version and kingpin? we are already using the version package from prometheus

mjtrangoni:
Hi @zwopir,

Ok thanks! I am good with this, and I will be fixing more issues later on.
What's about [dep](https://github.com/golang/dep)? I really like it, but it is up you to switch.
Should we open an issue for moving to kingpin?
If yes, we could move to something like the `node_exporter` does with version and handlers. See [here](https://github.com/prometheus/node_exporter/blob/master/node_exporter.go#L78).

zwopir:
I'm fine with dep. Had some issues on larger projects with it, but you manage to migrate to it, the project will benefit from it as there are probably more dep users out there than govendor.

The kingpin version handler looks good. Creating an issue for it and the kingpin migration is a good idea

PR #168, opened 2018-07-23
Add filesystem IO metrics

seanlaff:
Filesystem IO metrics are in the elasticsearch nodestats output, but they were not being exported.

In addition to the new IO metrics, the filesystem `data` metrics have had their variables renamed to avoid confusion. In elastic <5 the only metric within the filesystem output of nodeStats  was `data`, which is why I assume the naming was less specific. In elastic >5 the filesystem output of nodeStats contains both `data` as well as `io_stats`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,103 +58,108 @@ elasticsearch_exporter --help
 
 ### Metrics
 
-|Name                                                        |Type       |Cardinality   |Help
-|----                                                        |----       |-----------   |----
-| elasticsearch_breakers_estimated_size_bytes                | gauge     | 4            | Estimated size in bytes of breaker
-| elasticsearch_breakers_limit_size_bytes                    | gauge     | 4            | Limit size in bytes for breaker
-| elasticsearch_breakers_tripped                             | counter   | 4            | tripped for breaker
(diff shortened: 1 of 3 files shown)


zwopir:
Hi @seanlaff,

thanks for your contribution! Looks good. Can you fix the conflict so I can merge your PR?!

caarlos0:
I usually run node_exporter on the ES nodes as well... isn't it better to just use the node_exporter IO-related metrics?

zwopir:
I agree the node_exporter provides the more detailed set of metrics. The argument for having os related metrics in the exporter is mainly that you are usually not able to install the node_exporter in hosted es (by was or elastic for example)

caarlos0:
> I agree the node_exporter provides the more detailed set of metrics. The argument for having os related metrics in the exporter is mainly that you are usually not able to install the node_exporter on hosted es (by was or elastic for example)

hmm, that makes sense indeed.

seanlaff:
@zwopir Resolved the conflicts. Thanks for taking a look

zwopir:
thanks again for your contribution!

PR #176, opened 2018-08-18
add cluster label retriever

zwopir:
this is a work in progress PR to add common labels to all collectors. Not ready to merge, feedback welcome.

### DIFF
--- a/pkg/clusterinfo/clusterinfo.go
+++ b/pkg/clusterinfo/clusterinfo.go
@@ -15,6 +15,11 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+const (
+	namespace = "elasticsearch"
+	subsystem = "custerinfo"
+)
+
 var (
 	// ErrConsumerAlreadyRegistered is returned if a consumer is already registered
 	ErrConsumerAlreadyRegistered = errors.New("consumer already registered")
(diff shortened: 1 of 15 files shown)


zwopir:
Hi @dannyk81 and @Mikeschuette,

so the basic functionality is there. The PR just needs some more resilience checking + code style love. If you have some time, I'd appreciate your feedback on the solution

dannyk81:
dear @zwopir, apologies for the delay! I was away for a while.

Would love to give this a try, is this included in the latest RC? or should I build from this branch?

zwopir:
Hey @dannyk81, 

the feature is not yet included in a RC release nor merged into master. So you'd need to build it from the branch. 

There are also some question I'd love to discuss: if the ES node specified in the `es.uri` parameter isn't available at startup of the exporter, the startup is blocked in the first clusterlabel retrieval for at max 10s. If by that time no clusterlabel could be retrieved, a default label is returned. That label is later on (if the ES is available) replaced with the correct label. Unfortunately the "later on" can be anything between 0s and `-es.clusterinfo.interval`.
Does this make sense to you?

dannyk81:
@zwopir this sounds to me like a perfectly sane approach.

I'm doing some driver tests with this branch in a test cluster, will report.

zwopir:
@metalmatze can I ask for a review?

pkg/clusterinfo/clusterinfo.go
```diff
@@ -0,0 +1,276 @@
...
+	res, err := r.client.Get(u.String())
+	if err != nil {
+		_ = level.Error(r.logger).Log(
+			"msg", "failed to get cluster info",
+			"err", err,
+		)
+		return nil, err
+	}
+
+	defer func() {
```
metalmatze: I think it's ok to simply do `defer res.Body.Close()`

zwopir: yes, I totally agree. This clumsy construction is a victim of the newly introduced errcheck, which I consider as useful most of the time. I already had quite some "sh*t, errcheck and logging don't like each other", so I'm consider ditching errcheck again. Thoughts are welcome...

zwopir: and ... thanks for the review!

mjtrangoni: @zwopir what is about using Prometheus common/log library? I think, this would handle your logging problem better.

zwopir: it's beyond the scope of this PR, but maybe it's worth considering. I personally don't really like logrus (which prometheus logging is based on), which of cause shouldn't prevent us changing the logging if needed.

metalmatze: Prometheus itself uses the go-kit logger these days.

metalmatze: @zwopir Take a look at this example which turns of parts of errcheck:
https://github.com/improbable-eng/thanos/blob/master/.errcheck_excludes.txt

zwopir: thanks! I will definitely add errcheck excludes for logging and probably resp.body.Close(). You're doomed anyways if those operations fail >:(

PR #178, opened 2018-08-18
vendor github.com/blang/semver

marthjod:
add missing dep


zwopir:
Yeah, forgot to vendor semver, but then decided to vendor once the PR is ready. But there probably won't be any further dependency addition, so thanks + merge 😊

PR #179, opened 2018-08-18
green tests

marthjod:


pkg/clusterinfo/clusterinfo.go
```diff
@@ -68,36 +68,33 @@ func (r *Retriever) RegisterConsumer(c consumer) error {
 func (r *Retriever) Run(ctx context.Context) {
 	// start update routine
 	go func(ctx context.Context) {
-		for {
-			select {
-			case <-r.sync:
-				level.Info(r.logger).Log(
-					"msg", "providing consumers with updated cluster info label",
+		for range r.sync {
```
zwopir: might have to change that back to `for { select {} }` loop as the graceful loop termination on ctx cancellation is missing.

Issue #161, opened 2018-07-17
snapshots/repositories metrics


sp-francisco-manas:
Hi,

We're using this exporter with excellent results, thanks for the great work!

The only thing we're missing are snapshots/repositories metrics so we can add alerts in case some snapshot has failed or it's not being done.

The exporter will need to query another URLs to get the info:
- repository list: http://localhost:9200/_cat/repositories
- repository info (snapshots list): http://localhost:9200/_cat/snapshots/repository

If you think it makes sense to have this metrics, please let me know so we can work on a PR.

Cheers

zwopir:
Hi Francisco,

having metrics about snapshots certainly makes sense. We currently retrieve those with a separate gcp-specific, non-opensource exporter, but having a more generalized snapshot metrics in the exporter would be nice. If you feel confident creating you can create a PR and I will happily review it - after the weekend, as I'm currently on holidays with my kids. If you need help, I can help you alter the weekend as well

zwopir:
Hi @sp-francisco-manas,

I'm back from holidays. Are you working on a PR? do you need help?

sp-francisco-manas:
Hi @zwopir 

Sorry for the late reply, i've been on holidays too. I'm working on a PR, i hope to have it soon

Thanks

PR #182, opened 2018-08-28
initial snapshots stats

sp-joseluis-ledesma:
Hi,

  I took the work sp-francisco-manas was doing, so this PR solves #161

  There are a couple of things to check, and it lacks tests, but I would like to know if you find it misses something, or if there is something you would like to make different before we continue working on this.

regards,

### DIFF
--- a/collector/snapshots.go
+++ b/collector/snapshots.go
@@ -1,0 +1,297 @@
+package collector
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"path"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type snapshotMetric struct {
+	Type   prometheus.ValueType
+	Desc   *prometheus.Desc
+	Value  func(snapshotStats SnapshotStatDataResponse) float64
+	Labels func(repositoryName string, snapshotStats SnapshotStatDataResponse) []string
+}
+
+type repositoryMetric struct {
+	Type   prometheus.ValueType
+	Desc   *prometheus.Desc
+	Value  func(snapshotsStats SnapshotStatsResponse) float64
+	Labels func(repositoryName string) []string
+}
+
+var (
+	defaultSnapshotLabels      = []string{"repository", "state", "version"}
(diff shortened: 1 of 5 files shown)


collector/snapshots.go
```diff
@@ -0,0 +1,302 @@
...
+	Value  func(snapshotsStats SnapshotStatsResponse) float64
+	Labels func(repositoryName string) []string
+}
+
+var (
+	defaultSnapshotLabels      = []string{"repository", "state", "version"}
+	defaultSnapshotLabelValues = func(repositoryName string, snapshotStats SnapshotStatDataResponse) []string {
+		return []string{repositoryName, snapshotStats.State, snapshotStats.Version}
+	}
+	defaultSnapshotShardLabels = append(defaultSnapshotLabels, "metric")
```
zwopir: somehow `metric` is a bit to generic, but I'm struggling to find a better name. This is usually not a problem of the one who's trying to find the label name but a problem of the label itself. The label values simply have too less in common. So we should separate the metrics in
- `snapshot_total_shards`
- `snapshot_failed_shards`
- `snapshot_successful_shards`

Another hint for label design is, that it should make sense to sum over the labels. If it doesn't make sense, it shouldn't be in one metric.

Sorry for coming up with this a bit late.

I'd also suggest to drop the `last_` from all the snapshot metrics. Prometheus always scrapes the latest available data, so this seems a bit redundant

sp-joseluis-ledesma:
Hi!

  your comments make sense. I just updated the commit adding also all the missing documentation. I think the only missing part is the tests & travis, right?

regards,

mjtrangoni:
@sp-joseluis-ledesma you can check why Travis is failing locally with `make gometalinter`.

sp-joseluis-ledesma:
If you find OK what I've done I will start with the tests, time to launch a bunch of containers :)

zwopir:
yes, looks good! Some testing would be great

sp-joseluis-ledesma:
I've added tests for the snapshot stats. If there is no further requests from your side, this PR is ready to merge

sp-joseluis-ledesma:
Hi!
  just pushed the change you requested.

zwopir:
thanks a lot for your contribution!

PR #186, opened 2018-09-28
Fix refresh explanation in readme

neufeldtech:
Please correct me if I'm wrong, but it appears that these two stats were backwards in the README


zwopir:
sorry, I missed the PR. Thank you very much

PR #191, opened 2018-10-26
Fixed createTLSConfig function

username1366:
Return full TLS configuration when ca, cert, key, and insecure flag are set. 
Related to this issue: https://github.com/justwatchcom/elasticsearch_exporter/issues/158 
Tested on elasticsearch:6.4.2 + search-guard-6:6.4.2-23.1 TLS plugin


zwopir:
Hi @username1366,

thanks for the PR, looks good

PR #193, opened 2018-11-08
Fix codespell issues

mjtrangoni:
Hi @zwopir,

I fixed some trivial spelling issues. See,
````
$ codespell -S ./vendor*,./.git*
````

Would you consider adding this to CircleCI, so that it will be always checked?


zwopir:
Hi Mario, thanks for your contribution, looks good!

PR #194, opened 2018-11-09
Possibility to add Proxy to http.client Transport

czerwe:
By default this function will be skipped to avoid breaking changes.

Two new flags are added:
  -network.proxy   -   which can be the proxy host
  -network.proxy-enable  - if this should be used at all

If network.proxy is not passed the system tries to read the env variable http_proxy or https_proxy. Which of both he tries to read will be decided due to the prefix of the uri (e.g. https -> https_proxy)

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Below is the command line options summary:
 elasticsearch_exporter --help
 ```
 
+
 | Argument                | Description | ENV_VAR | Default     |
 | --------                | ----------- | ----------- |----------- |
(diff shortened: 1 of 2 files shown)


zwopir:
Hi @czerwe,

thanks for your PR. As far as I understand the documentation, the default transport already supports specifying a proxy via environment variable. See https://godoc.org/net/http#RoundTripper. Unless I'm wrong here, it doesn't make sense to me to add your contribution to the exporter :(

czerwe:
Hi @zwopir,

I verified it now. That's partly correct and my pull request should not be merged.

If I change the following code inside the main.go:

	httpClient := &http.Client{
		Timeout: *esTimeout,
		Transport: &http.Transport{
			TLSClientConfig: tlsConfig,
		},
	}

with
	httpClient := &http.Client{
		Timeout: *esTimeout,
		Transport: &http.Transport{
			TLSClientConfig: tlsConfig,
			Proxy:           http.ProxyFromEnvironment,
		},
	}

Then it's working. The current implementation of the http.Transport does not do it.
Please close this merge request. Should I open an other one with the "tiny" change?

PR #196, opened 2018-11-13
add collector for cluster settings

TheMeier:
add optional scraping of cluster settings (`cluster.routing.allocation.enabled` only so far)


zwopir:
Hi @TheMeier,

thanks for your contribution! The code looks fine, except for some file misspellings (culster vs cluster). So far I don't understand the use case for the exported metric(s). Can you elaborate a bit on it please?

TheMeier:
The use case is to detect wether the cluster wide shard allocation is in the desired  state or in an undesired state due to reasons like the following: 

a) cluster shard allocation was automatically set to none because `cluster.routing.allocation.disk.watermark.flood_stage` was reached
b) manually changing the allocation for e.g. maintenance and forgetting to switch it back
c) interrupted automated maintenance which changed the allocation and did not clean up because of the interruption


Fixed the filenames  :)

zwopir:
thanks for the explanations. It makes sense to have the possibility to check for the above conditions

zwopir:
Can you add the command line flag to the `README.md`, I'll then merge the PR

TheMeier:
updated README.md and fixed commit author

PR #200, opened 2018-11-30
expose name and date together with "index_name-date" 

joaopereiramrf:
As we want to do some aggregations by name and date of index, we need to expose name and date together with the format "index_name-date".

example:
elasticsearch_index_stats_fielddata_evictions_total{index=".kibana-20180630",**indexDate="20180630"**,**indexPrefix=".kibana"**} 0

### DIFF
--- a/collector/indices.go
+++ b/collector/indices.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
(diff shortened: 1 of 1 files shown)


zwopir:
as explained in #201 this PR doesn't make sense for most users

PR #203, opened 2018-12-09
Allow Comma separated list of multiple ES servers

### LINKED ISSUE #117: Allow Comma separated list of multiple ES servers
Hi there,

Thanks for this great exporter. I have been struggling to get the Prometheus exporter which supports older version i mean < 2.3.1. This works great. 

Currently functionality only allows single elastic server instance -es.uri=http:/serverA:9200.

It would be great if we can pass multiple server endpoints like 
-es.uri=http:/serverA:9200, serverB:9200,serverC:9200,

This way, even if Server A goes down, exporter still able to collect the data from the second server.

Misterhex:
Fix #117, Allow Comma separated list of multiple ES servers.

Allow Comma separated list of multiple ES servers for high availability scenarios. This allow running a single exporter container/process rather than N number of container/process per elasticsearch servers.

Before fetching stats from api, does http HEAD requests to list of es servers, pick the first url that responded, and use the healthly url to fetch stats from api.

### DIFF
--- a/collector/cluster_health.go
+++ b/collector/cluster_health.go
@@ -36,9 +36,10 @@ type clusterHealthStatusMetric struct {
 
 // ClusterHealth type defines the collector struct
 type ClusterHealth struct {
-	logger log.Logger
-	client *http.Client
-	url    *url.URL
+	logger    log.Logger
+	client    *http.Client
+	urls      []*url.URL
+	httpProbe HTTPProbe
 
 	up                              prometheus.Gauge
 	totalScrapes, jsonParseFailures prometheus.Counter
@@ -48,13 +49,14 @@ type ClusterHealth struct {
 }
 
 // NewClusterHealth returns a new Collector exposing ClusterHealth stats.
-func NewClusterHealth(logger log.Logger, client *http.Client, url *url.URL) *ClusterHealth {
+func NewClusterHealth(logger log.Logger, client *http.Client, urls []*url.URL) *ClusterHealth {
 	subsystem := "cluster_health"
 
 	return &ClusterHealth{
-		logger: logger,
-		client: client,
-		url:    url,
+		logger:    logger,
+		client:    client,
+		urls:      urls,
+		httpProbe: NewHTTPProbe(urls, client),
 
 		up: prometheus.NewGauge(prometheus.GaugeOpts{
 			Name: prometheus.BuildFQName(namespace, subsystem, "up"),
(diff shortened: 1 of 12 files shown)


zwopir:
Hi @Misterhex,

thanks for the PR! I'm currently on holidays and won't make it until Mon, 17.12. to review it. Maybe @metalmatze can have a look

Misterhex:
@zwopir thanks

TheMeier:
IMHO this is an anti-pattern. One exporter should be responsible for exactly one service instance.

Misterhex:
If we enable stats for all nodes, scrapping whichever nodes in the cluster pretty give the same stats. 

The api endpoint of elasticsearch gives the same result no matter which node we query it from.

We notice this as a problem when our 1 of the node went down but the cluster is still healthly but could have been mitigited if we give a couple more uris to the exporter.

metalmatze:
Both should work for Prometheus. Prometheus itself doesn't really care. But alerting on a down exporter is way easier when having multiple exporter rather than a big one.
I personally wouldn't be too opinionated about it, if there are reason for running a bigger one, but in general would prefer multiple ones.

zwopir:
I also don't have a strong opinion about it. On the pro big-exporter side I see the use case of a hosted ES cluster. On the other hand I'm not really happy about the doubled http calls (HEAD plus GET) on each scrape and collector. The more or less official go ES client has a stateful connection object that manages the connections to the cluster nodes. It has a sophisticated node sniffing logic and in band connection checking. This seems the right way to me of doing it, but it's way too much logic for a simple exporter

Misterhex:
I guess our use case is slightly different, as we are targeting elasticsearch cluster running external to kubernetes, so having N pods matching to N external nodes seems wasteful.

If we were running elasticsearch on kube, I would imagine a exporter sidecar contaier for each es pod. But without kube, this is not really feasible.

Please feel free to close the PR, thanks for reviewing!

metalmatze:
Agreed @Misterhex. I can see your use case being one for running a "bigger" exporter. For Kubernetes I still favor sidecars for this.

Misterhex:
Agreed @zwopir, closing the PR, agree the es client is the right "bigger" exporter if there are future plan to support this.

Misterhex:
For anyone interested, the solution i adopted was by problems/concerns to the network layer by running envoy as a side car proxy to the exporter. It does the forward load balacning and health check.

sirkubax:
> Both should work for Prometheus. Prometheus itself doesn't really care. But alerting on a down exporter is way easier when having multiple exporter rather than a big one.
> I personally wouldn't be too opinionated about it, if there are reason for running a bigger one, but in general would prefer multiple ones.

Prometheus may not care, but it is a nightmare to filter it out in grafana...
image
 irate(elasticsearch_jvm_gc_collection_seconds_count{cluster="$cluster",name=~"$name"}[$interval])


what would be your solution to remove duplicates? There is no 'distinct' filter in promql

Issue #207, opened 2018-12-14
Place the node role as a label value


jorgelbg:
> I'm opening this issue to discuss if this change would make sense

At the moment the role of each server in the cluster is stored in a label with a `true/false` value:

```
metric_name{cluster="cluster1",es_client_node="true",es_data_node="false",
es_ingest_node="false", es_master_node="true",...}
```

This provides the ability to filter by a given role in the cluster, but it's not very *friendly* to dashboard solutions.

Having the role as the value of a label (`...,role="client",...`) could be easier to integrate into some dashboarding solutions. For instance, on Grafana, this value could be fetched from the metric with the `label_values(metric, label)` function. This would allow using the "values group" feature.

So far the alternative that I've found for this is to accomplish this using relabeling. If the feature is welcomed I can submit a PR.

zwopir:
Hi @jorgelbg,

we deliberately decided to encode the role info as bool-style key/value pairs. So the proposed node can have multiple roles and a role label value could be for example `data,ingest,client`. Evaluating such a label in a prom-query (dashboard) would require the knowledge how multiple roles are encoded in the label value. With the actual solution you only need to know the label names. So from my point of view I'd rather not change the way we include the role information in the metrics.

What do you think, @marthjod , @metalmatze ?

jorgelbg:
@zwopir Thanks for the reply and for opening a discussion about this.

Regarding the multiple roles I've run into this scenario as well. I have some nodes with mixed roles in my clusters as well. Indeed, `es_*_node` label is very handy for filtering but makes extracting the possible roles programmatically quite complicated (unless they're hardcoded). On the other hand, having multiple roles on a single label gets tricky (as you said).

I was thinking about a similar approach to the one discussed here https://www.robustperception.io/how-to-have-labels-for-machine-roles. But perhaps this approach could cause some other issues 🤔.

zwopir:
that's actually a very nice solution to our problem. I wasn't of it. Thanks for bringing it up!
So yes, if you could contribute a PR, that would be very nice. It shouldn't be too much work as the logic of figuring out the node role is already in the code.

Later in the life-cycle if the exporter we may even think of deprecating the `es_*_node` label.

zwopir:
see #228

zwopir:
Hi @jorgelbg,

I implemented the feature in #228. Can you build from that branch and report back if it works for you?

jorgelbg:
[...]

PR #228, opened 2019-02-27
add node_roles metric (#207)

zwopir:


jorgelbg:
I've tested the exporter locally, what I'm realizing is that we're exposing 1 metric per role, per node. So for a node that it is only a client & ingest node we get 1 metric, with the 4 `role` labels.

```
elasticsearch_nodes_roles{cluster="nomnom",host="192.168.99.100",name="logparser3",role="client"} 1
elasticsearch_nodes_roles{cluster="nomnom",host="192.168.99.101",name="logparser3",role="data"} 0
elasticsearch_nodes_roles{cluster="nomnom",host="192.168.99.102",name="logparser3",role="ingest"} 1
elasticsearch_nodes_roles{cluster="nomnom",host="192.168.99.103",name="logparser3",role="master"} 0
```
I think that the ones with `0` value are not needed. An additional comparison (no label match) would be needed in the query to do the `group_left`. My initial POC was to handle the roles outside of `c.nodeMetrcs`, in it's own block. I've entangled my POC changes with your PR in: https://github.com/jorgelbg/elasticsearch_exporter/commit/db863464b9d5aebe5e6c46a0e54fa71ca4ae5361.

I've kept the number of changes to a minimum, to align with your PR. `createRoleMetric` doesn't need to be an instance of `nodeMetric`, it can return/initialize a custom struct, or it can be inlined in the `for` loop, since `getRoles` it is called only once, and not evaluated on a per-role basis.

zwopir:
yes, you're right. Thanks! I'll just pull your changes into `feature/role_metrics`

zwopir:
@jorgelbg can you test again?

jorgelbg:
I think that it's working as expected, this is the output for one of my client nodes:

```
elasticsearch_nodes_roles{cluster="nomnom",host="192.168.99.100",name="logparser3",role="client"} 1
elasticsearch_nodes_roles{cluster="nomnom",host="192.168.99.100",name="logparser3",role="ingest"} 1
```

Which matches our expectation, ingest & client node.

Issue #220, opened 2019-02-01
Add index_stats_search_scroll_current metric


adeesent:
Greetings!

As I can see, there is a such field in struct, but no metric exposed - https://github.com/justwatchcom/elasticsearch_exporter/blob/c36d0a21ea7e3195297e1de1a77fd1ce93103d30/collector/indices_response.go#L104

zwopir:
Hi @adeesent,

we happily accept PRs, especially if they don't introduce extra calls to ES API endpoints. I probably releasing a new release candidate within the next week, so if you're fast, the PR can make it into the next release. Let me know, if you need help

PR #232, opened 2019-03-02
Add search.scroll_current metric

adeesent:
#220

### DIFF
--- a/collector/indices.go
+++ b/collector/indices.go
@@ -501,6 +501,18 @@ func NewIndices(logger log.Logger, client *http.Client, url *url.URL, shards boo
 				Labels: indexLabels,
 			},
 			{
+				Type: prometheus.GaugeValue,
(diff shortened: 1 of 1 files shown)


collector/indices.go
```diff
@@ -500,6 +500,18 @@ func NewIndices(logger log.Logger, client *http.Client, url *url.URL, shards boo
 				},
 				Labels: indexLabels,
 			},
+			{
+				Type: prometheus.CounterValue,
```
zwopir: are you sure this is a counter? i.e. the value can only increase. If so, please add a `_total` suffix to the metric name. If this is not a counter, please change the type to `prometheus.GaugeValue`

adeesent: Yeah, you are right. It's definitely gauge. Thank you. Fixed!

zwopir:
thanks again

PR #242, opened 2019-05-02
Improve environment variable documentation

slovdahl:
From looking through the code, it seems like `-` in parameter names should be replaced by `_` as well.


zwopir:
that's right. Thanks for pointing out and fixing the documentation

PR #243, opened 2019-05-10
Add index_stats_query_cache_caches_total metric

kaorimatz:
This is the index-level version of the indices_query_cache_cache_count metric, which is the total number of cache entries that has been stored in the query cache.

### DIFF
--- a/collector/indices.go
+++ b/collector/indices.go
@@ -827,6 +827,18 @@ func NewIndices(logger log.Logger, client *http.Client, url *url.URL, shards boo
 			{
 				Type: prometheus.CounterValue,
 				Desc: prometheus.NewDesc(
+					prometheus.BuildFQName(namespace, "index_stats", "query_cache_caches_total"),
+					"Total query cache caches count",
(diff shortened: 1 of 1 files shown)


zwopir:
Hi @kaorimatz,

thanks for your contribution to the exporter. Are you sure the metric is a counter, i.e. a value that can only go up?

kaorimatz:
Hi @zwopir, yes `cache_count` is the number of cache entries that has been stored in the cache, which only goes up or is reset. `cache_size` is the current number of entries in the cache, which goes down on an eviction:

https://github.com/elastic/elasticsearch/blob/ed3230b3ebbddbe0cf049c76ce42fde79f1b793c/server/src/main/java/org/elasticsearch/indices/IndicesQueryCache.java#L305-L342
https://github.com/elastic/elasticsearch/blob/ed3230b3ebbddbe0cf049c76ce42fde79f1b793c/server/src/main/java/org/elasticsearch/index/cache/query/QueryCacheStats.java#L98-L110

PR #244, opened 2019-05-10
Change type for indices_query_cache_cache_count metric to counter

kaorimatz:
indices.query_cache.cache_count is the total number of cache entries that has been stored in the query cache, which should only increase or be reset.


zwopir:
thanks, are you sure it's a counter? I don't exactly know the meaning of the underlying json field. In case it gets adjusted on cache evictions, it's not a counter.

If you're sure about it, a `_total` is missing in the metric name (following the prometheus metric name conventions)

kaorimatz:
Yes, `cache_count` is a counter which gets incremented when an entry is stored in the cache and shouldn't get decremented on an eviction:

https://github.com/elastic/elasticsearch/blob/ed3230b3ebbddbe0cf049c76ce42fde79f1b793c/server/src/main/java/org/elasticsearch/indices/IndicesQueryCache.java#L305-L342
https://github.com/elastic/elasticsearch/blob/ed3230b3ebbddbe0cf049c76ce42fde79f1b793c/server/src/main/java/org/elasticsearch/index/cache/query/QueryCacheStats.java#L98-L103

PR #277, opened 2019-08-12
Add metric for cluster.max_shards_per_node

### LINKED ISSUE #271: Add metric for cluster.max_shards_per_node
In elasticsearch 7.x was introduced soft limit on the number of shards in a cluster https://www.elastic.co/guide/en/elasticsearch/reference/7.0/misc-cluster.html#cluster-shard-limit. As mentioned in the documentation reaching this limitation can lead to failures:
```
If the cluster is already over the limit, due to changes in node membership or setting changes, all operations that create or open indices will fail until either the limit is increased ..., or some indices are closed or deleted to bring the number of shards below the limit.
```
So it might be useful to have dashboard/alert to indicate shards usage before the actual failures.

TheMeier:
First draft, still need to do some testing. Also want to clean up and put the test json strings in separte fixture files.

fixes #271

Signed-off-by: Christoph Maser <christoph.maser+github@gmail.com>


zwopir:
nice, thanks a lot! Looks good so far. Ping me if you want to merge

TheMeier:
So, turns out i forgot to encode the query params and the `?include_defaults=true` was never send. This was a bug also for the existing shard allocation metric.

I also moved the test json content into files under a new `fixtures` directory. Probably we should consider a util package to make it easy to setup test strings from fixture files and then refactor the existing tests to use fixture files instead of inline json strings.

@zwopir BTW its quite easy to generate arbitrary test fixtures by simply firing up a container eg
```
docker run --rm -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:7.3.0
```
And then fetching the data, eg: 
```
curl "http://localhost:9200/_cluster/settings?include_defaults=true" | jq .> fixtures/settings-7.3.0.json
```

@zwopir so i think this is ready to review :)

collector/cluster_settings_test.go
```diff
@@ -14,10 +17,13 @@ func TestClusterSettingsStats(t *testing.T) {
...
-	for ver, out := range tcs {
+	files := []string{"../fixtures/settings-5.4.2.json", "../fixtures/settings-merge-5.4.2.json"}
+	buf := bytes.NewBuffer(nil)
```
zwopir: there's no need to use a buffer. You can `io.Copy(w,f)` in the http handler

collector/cluster_settings_test.go
```diff
@@ -35,10 +41,48 @@ func TestClusterSettingsStats(t *testing.T) {
...
+		}
+	}
+}
+
+func TestClusterMaxShardsPerNode(t *testing.T) {
+	// Testcases created using:
+	//  docker run -d -p 9200:9200 elasticsearch:VERSION-alpine
+	//  curl http://localhost:9200/_cluster/settings/?include_defaults=true
+	files := []string{"../fixtures/settings-7.3.0.json"}
+	buf := bytes.NewBuffer(nil)
```
zwopir: same here

TheMeier: Hm ok, still smells like a case of don't repeat yourself to me but if you can live with that... ;)

zwopir:
thank you!

tetianakravchenko:
Any chance to get it released?

KevW7:
Hi, I'm really interested in this functionality - do you have any plans to create a new release?

PR #291, opened 2019-10-03
Fix spelling mistake in shards docs metric name

shaunmansell:
Updating the shards docs metric name to fix spelling mistake. Final metric name will be `elasticsearch_indices_shards_docs`.

Fixes https://github.com/justwatchcom/elasticsearch_exporter/issues/290


zwopir:
thanks!

Issue #156, opened 2018-06-12
more shard-related metrics


caarlos0:
It would be nice if we could have metrics like:

- number of replica shards of a given index
- shard size

I'm pretty new to ES, but where it seems important to track the shard size to decide if it is a good idea to reindex with more shards for example. 

The number of replicas for a given index may be important to alert if there are too few replicas - which could mean that a data loss may happen.

If you agree this information is valuable, I can work on PR :) 

Just wanted to check first...

Thanks again for the great project BTW 🍻

zwopir:
Hi Carlos,

yes, it sounds reasonable to track the shard metrics, but not critical. The elasticsearch_exporter exports the cluster status (red, yellow, green) which includes the information about "under-replicated" shards. 

So if the metrics are already in the JSON we get from querying ES, I'm happy to accept your PR. If not or if we had to add new parameters to the curl, thus increasing the size of the transferred JSON or adding new curls to the ES, let me think twice.

Thanks for offering on the PR, I'll happily review it!

charles-mims-sp:
Shard size is what I came here looking for. When shards get around 50GB or so, performance starts to really suffer.

elwood218:
Hi there. Yes it would also be nice if we could habe the `number_of_shards` setting per index. This can be used for calculating things.
Furthermore I would be interested in that Merge what is mentioned here about `shards_store_size_in_bytes`. But there is no Tag and also no Docker Image build of that which included this new metric or am I missing something?!

caarlos0:
You can use `caarlos0/elasticsearch_exporter:1.2.0-alpha11`, it has more things though: https://github.com/justwatchcom/elasticsearch_exporter/compare/master...caarlos0:dev?expand=1

PR #292, opened 2019-10-04
added shard store size as well as "primary" label to shard metrics

macgyver603:
refs #156 

I added `shards_store_size_in_bytes` as one of the shard-level metrics exposed through the exporter.

I also added a `primaries` label to the shard-level metrics in order to differentiate between primaries and replicas.

As for counting replica shards, there isn't a specific metric for tracking that, but a quick 'n dirty way of getting that would be to count the number of of occurrences of a specific metric for a specific shard with something like:

```
count(elasticsearch_indices_shards_docs_deleted{index="my-index", shard="0"})
```

### DIFF
--- a/collector/indices.go
+++ b/collector/indices.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strconv"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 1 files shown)


zwopir:
thanks for your PR, looks good. Merged

PR #300, opened 2019-11-05
added missing data sources

mastak:
Added missing data sources in `dashboard.json` example


mumrau:
Bumping this up, I faced a similar issue.

sysadmind:
The dashboard has been replaced and the new one does include the $datasource var for metrics

PR #308, opened 2019-12-12
Allow passing credentials through env variables

malletgu:
This MR will allow to inject credentials through environment variables. Also, if credentials are passed in the URI and as arguments, the arguments will overwrite the one in the URI

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,8 @@ elasticsearch_exporter --help
 | es.client-cert          | 1.0.2                 | Path to PEM file that contains the corresponding cert for the private key to connect to Elasticsearch. | |
 | es.clusterinfo.interval | 1.1.0rc1              |  Cluster info update interval for the cluster label | 5m |
 | es.ssl-skip-verify      | 1.0.4rc1              | Skip SSL verification when connecting to Elasticsearch. | false |
+| es.username             | 1.2.0                 | Username to use for connecting to Elasticsearch | false |
+| es.password             | 1.2.0                 | Password to use for connecting to Elasticsearch | false |
 | web.listen-address      | 1.0.2                 | Address to listen on for web interface and telemetry. | :9114 |
 | web.telemetry-path      | 1.0.2                 | Path under which to expose metrics. | /metrics |
 | version                 | 1.0.2                 | Show version info on stdout and exit. | |
@@ -65,6 +67,9 @@ For versions greater than `1.1.0rc1`, commandline parameters are specified with 
 by replacing `.` and `-` with `_` and upper-casing the parameter name.
 
 #### Elasticsearch 7.x security privileges
+
+Username and password can be passed either directly in the URI or through the arguments `es.username` and `es.password`.
+Specifying those two arguments will override authentication passed in the URI (if any).
 
 ES 7.x supports RBACs. The following security privileges are required for the elasticsearch_exporter.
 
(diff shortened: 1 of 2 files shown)


pkoraca:
Hi, any chance this might be accepted and released soon? Thanks

malletgu:
> Hi, any chance this might be accepted and released soon? Thanks

I know it is not ideal but if you're interested I made the Docker in my personal docker hub in the meantime: https://hub.docker.com/r/malletgu/elasticsearch-exporter/tags

jackatbancast:
Just want to bump this. As @malletgu mentioned, anyone using the operator from @elastic would have to fork currently, as you can't build an environment variable dynamically in kubernetes.

It would be good to get this merged so that everyone can benefit from any upstream changes.

jcruzmartini:
It would be really nice to have this one released soon. I am using K8 and I can not create dynamic env vars
Thanks

sysadmind:
I was able to use this PR to get metrics from an [ECK elasticsearch cluster](https://www.elastic.co/guide/en/cloud-on-k8s/current/index.html). I can confirm that this works when specifying the URl (pointing to the kubernetes service) and setting the credentials through the variables. 

One thing to note is that I was not able to use the secret that ECK created directly with the helm/stable chart. The secret "key" is set to the user name and the value is the password. You can do something like the following though (maybe this needs to be a PR to the helm chart).
```
env:
- name: ES_USERNAME
  value: "elastic"
- name: ES_PASSWORD
    valueFrom:
    secretKeyRef:
        name: my-cluster-es-elastic-user
        key: elastic
```

ambis:
Just stumbeled on this problem and this PR seems to solve it nicely! Waiting for merge...

schrej:
While this is convenient it's already possible to do this, see [my comment](https://github.com/justwatchcom/elasticsearch_exporter/issues/307#issuecomment-623595480) on the issue.

kfox1111:
+1 to getting this functionality merged. Having to create a new secret and keep it synchronized is not as good as referencing the secret in an existing secret.

Glad there is a workable workaround though.

kfox1111:
Actually, it looks like you can use 
extraEnvSecrets:
  ES_PASSWORD:
    secret: foo-elastic-user
    key: elasticsearch

to do the job...

So the chart does work as is.

Maybe an example might help? The configuration can be made to work as is, but its not very clear how to do it.

pazoozooCH:
+1

smforsberg:
I would definitely encourage this being merged. The solution in the other case is fragile as it requires a strict ordering of multiple blocks of a helm chart interacting with its service. We can already see the example using `extraEnvSecrets` failing when several has implemented it (thus requiring the `envFromSecret` method). This will almost guaranteed cause a future breakage.

As this utility/exporter is for only elastic, it makes sense to support the elastic authorization methods which includes username/password without having to code along with URI (which makes secret management but discoverability of remote connection) more difficult.

SuperQ (changes requested):
This change allows setting the password on the command line as a flag. This is unsafe and we should not support this.

guillaumelecerf:
Maybe this approach would be acceptable https://github.com/prometheus-community/elasticsearch_exporter/pull/461 ?

guillaumelecerf:
https://github.com/prometheus-community/elasticsearch_exporter/pull/461 is now merged.

PR #318, opened 2020-01-22
feat: latest snapshot timestamp metric

caarlos0:
added a new metric that reports the timestamp of the latest `SUCCESS`/`PARTIAL` snapshot.

This should make it easier to alert things like `X time since the last snapshot`, even more helpful when you have some sort of repository rotation (e.g. each month has a new repository / s3 bucket, and we keep up to 2 old repositories there).

with this metric the alert rule could be:

```
(time() - max without (repository) elasticsearch_snapshot_stats_latest_snapshot_timestamp) > X
```

Thanks!

### DIFF
--- a/collector/snapshots.go
+++ b/collector/snapshots.go
@@ -186,7 +186,7 @@ func NewSnapshots(logger log.Logger, client *http.Client, url *url.URL) *Snapsho
 			{
 				Type: prometheus.GaugeValue,
 				Desc: prometheus.NewDesc(
-					prometheus.BuildFQName(namespace, "snapshot_stats", "latest_snapshot_timestamp"),
+					prometheus.BuildFQName(namespace, "snapshot_stats", "latest_snapshot_seconds"),
 					"Timestamp of the latest SUCCESS or PARTIAL snapshot",
 					defaultSnapshotRepositoryLabels, nil,
 				),
--- a/collector/snapshots.go
+++ b/collector/snapshots.go
@@ -183,6 +183,28 @@ func NewSnapshots(logger log.Logger, client *http.Client, url *url.URL) *Snapsho
 				},
 				Labels: defaultSnapshotRepositoryLabelValues,
 			},
+			{
+				Type: prometheus.GaugeValue,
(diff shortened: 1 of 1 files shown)


collector/snapshots.go
```diff
@@ -183,6 +183,24 @@ func NewSnapshots(logger log.Logger, client *http.Client, url *url.URL) *Snapsho
 				},
 				Labels: defaultSnapshotRepositoryLabelValues,
 			},
+			{
+				Type: prometheus.GaugeValue,
+				Desc: prometheus.NewDesc(
+					prometheus.BuildFQName(namespace, "snapshot_stats", "latest_snapshot_timestamp"),
```
grobie: It'd be great to follow Prometheus best practices and include the metric unit in description and metric name, like "Timestamp of the latest SUCCESS or PARTIAL snapshot in seconds" and `...latest_snapshot_timestamp_seconds`. See https://prometheus.io/docs/practices/naming/

caarlos0: agreed, I named it like this because the other metric is `oldest_snapshot_timestamp` already, so I guess its best to have all metrics here in the same pattern ? 🤔

anyway, I'm happy to change it if maintainers think we should...

zwopir: yes, please change the metric names so they follow the best practise. I'll create a new release after the merge and add a note to the release notes about the breakting change.

Thanks for your contribution and sorry I'm a bit slow in reacting to PR and issues here

caarlos0:
```suggestion
					prometheus.BuildFQName(namespace, "snapshot_stats", "latest_snapshot_seconds"),
```

caarlos0:
done.

caarlos0:
@zwopir done

grobie:
Great, thanks @caarlos0! I'm happy to see the explicit names, with tab completion in prometheus and grafana, abbreviations should be rather avoided.

caarlos0:
Me too 🙏

zwopir:
thanks again for the PR, @caarlos0 !

PR #321, opened 2020-02-13
Add path nodes collector params support

ernestas-poskus:
Thank you for this package :+1: . Unfortunately, it crashes when used on over 550 nodes or with over 90k active_shards with `--es.all` collection enabled. Obviously it is cheaper to scrape 1 to 3 nodes than 550. Therefore adding a few changes:

First https://github.com/justwatchcom/elasticsearch_exporter/commit/3b77b70688bebd1d6e8671808970565ec4bbaa8d change adds additional path support to /_nodes/stats.
Path addition allows filtering of relevant metrics, e.g.:

```bash
--es.stats.params=adaptive_selection,discovery,fs,http,indices,jvm,os,process,thread_pool,transport
```

This allows having multiple elasticsearch exporters that only return relevant /_nodes/stats:
 - `--es.stats.params=fs`
 - `--es.stats.params=http`
 - etc.

Second https://github.com/justwatchcom/elasticsearch_exporter/commit/a8fc87baa775483ab7df9949b8ed5a9af6cb1ec7 change controls whether to collect `/_nodes/stats` at all.

Third https://github.com/justwatchcom/elasticsearch_exporter/commit/ff1c7ce518d80c56de08c0430af4f3f65485406b change controls whether to collect `/_cluster/health` at all.

### DIFF
--- a/collector/nodes.go
+++ b/collector/nodes.go
@@ -156,6 +156,7 @@ type Nodes struct {
 	url    *url.URL
 	all    bool
 	node   string
+	params string
 
 	up                              prometheus.Gauge
 	totalScrapes, jsonParseFailures prometheus.Counter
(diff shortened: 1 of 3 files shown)


ArthurSens:
Hi 👋 -- Sorry for the super-delayed review here. Unfortunately, the codebase has changed a lot in the last 6 years, and the current implementation doesn't make much sense anymore.

The codebase is currently being refactored, moving metric generation to a package we call `collectors`. With collectors, one can enable/disable them with CLI flags `--collector.<collector-name>`. Unfortunately, nodes and cluster metrics haven't been refactored yet, but that refactor should solve the same problem as this PR.

I'm taking the liberty of closing this one, but we definitely need help with the remaining refactoring!

PR #324, opened 2020-02-20
fix metric name for gc counter

dpavlos:
Hello and thanks for this great exporter! I have noticed that there is a typo in `elasticsearch_jvm_gc_collection_seconds_count` metric which actually counts gc events and not seconds.


dpavlos:
Any chance of reviewing this? @zwopir @pakohan  Thanks in advance!

sysadmind (approved):
LGTM

Issue #329, opened 2020-03-27
Support fo go modules


badrabubker:
Hi, 
thanks for providing elasticsearch-exporter. I was trying to package this exporter for  [debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=935317).

First issue i had is that  [prometheus-client-golang](https://github.com/prometheus/client_golang) in the vendor directory is outdated and newer versions [break computability](https://github.com/prometheus/client_golang/blob/master/CHANGELOG.md#100--2019-06-15) with the < 1 versions .
So i [patched](https://salsa.debian.org/go-team/packages/prometheus-elasticsearch-exporter/-/commit/eac41bef6951a50f516a19da2e14c5af44005945) that in the package.
I tired to vendor it as well, however the newer [prometheus-client-golang](https://github.com/prometheus/client_golang) depends on **github.com/cespare/xxhash/v2** which needs go modules

I would be happy to submit a pull request if you would like :) 

Thanks ! 
Cheers,
Badr

PR #333, opened 2020-04-07
Add go module support and deprecate vendored deps

sysadmind:
Adds go module support (go.mod, go.sum) and removes the vendor directory.

This should address #329 including the update to the prometheus/client_golang dependency to a version that supports go modules.

I am certainly no make expert so this may not be 100% but I was able to run the tests and the build along with docker and all of them were successful. I feel relatively certain with the existing tests that this should not break anything for existing users. 

Also, thank you for the work put in to developing and maintaining this project. It is a huge help.


sysadmind:
closing in favor of #350 
Will post the few differences that I have here over on that PR.

PR #338, opened 2020-04-13
feat: snapshots interval

caarlos0:
We have snapshots on GCS, and gathering snapshots metrics takes ~2min sometimes, causing the whole thing to timeout often.

My idea with this one is to get snapshots metrics only every now and then, and do so in background.

The old behavior is still the default, and this new behavior is opt-in.

what do you think?

### DIFF
--- a/collector/snapshots.go
+++ b/collector/snapshots.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
@@ -43,6 +44,11 @@ type Snapshots struct {
 	client *http.Client
 	url    *url.URL
 
+	updateInterval time.Duration
+	lastResponse   map[string]SnapshotStatsResponse
+	lastError      error
+	lastFetchAt    time.Time
+
 	up                              prometheus.Gauge
 	totalScrapes, jsonParseFailures prometheus.Counter
 
@@ -51,11 +57,12 @@ type Snapshots struct {
 }
 
 // NewSnapshots defines Snapshots Prometheus metrics
-func NewSnapshots(logger log.Logger, client *http.Client, url *url.URL) *Snapshots {
+func NewSnapshots(logger log.Logger, client *http.Client, url *url.URL, interval time.Duration) *Snapshots {
 	return &Snapshots{
-		logger: logger,
-		client: client,
-		url:    url,
+		logger:         logger,
+		client:         client,
+		url:            url,
+		updateInterval: interval,
 
 		up: prometheus.NewGauge(prometheus.GaugeOpts{
 			Name: prometheus.BuildFQName(namespace, "snapshot_stats", "up"),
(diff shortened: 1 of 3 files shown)


collector/snapshots.go
```diff
@@ -276,20 +283,28 @@ func (s *Snapshots) Collect(ch chan<- prometheus.Metric) {
...
-	// indices
-	snapshotsStatsResp, err := s.fetchAndDecodeSnapshotsStats()
-	if err != nil {
+	if s.updateInterval == 0 {
+		_ = level.Debug(s.logger).Log("msg", "getting snapshots metrics", "block", "true")
+		s.lastResponse, s.lastError = s.fetchAndDecodeSnapshotsStats()
+	} else if s.lastFetchAt.Add(s.updateInterval).Before(time.Now()) {
+		go func() {
+			_ = level.Debug(s.logger).Log("msg", "getting snapshots metrics", "block", "false")
+			s.lastFetchAt = time.Now()
```
sysadmind, line 292: [According to the docs](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus?tab=doc#Collector), the `Collect()` method may be called concurrently so it should be  concurrency safe which I do not believe these additions adhere to. `lastFetchAt`, `lastResponse`, and `lastErrror` can all change during the lifetime of this method call. I think there needs to be some sort of mutex to handle the concurrency or there may be a way to refactor in order to avoid the need for a mutex.

caarlos0:
Did some improvements:

- singleflight to avoid multiple calls concurrent calls to the snapshot APIs (which, as said before, are slow when the repository is remote)
- mutex to lock concurrent writes to `s.lastUpdateAt` et al

collector/snapshots.go
caarlos0: done, thanks for the feedback :)

caarlos0:
cc/ @zwopir

collector/snapshots.go
```diff
@@ -43,6 +46,13 @@ type Snapshots struct {
 	client *http.Client
 	url    *url.URL
 
+	updateInterval time.Duration
+	lastResponse   map[string]SnapshotStatsResponse
+	lastError      error
+	lastFetchAt    time.Time
+	lock           sync.RWMutex
+	group          singleflight.Group
```
sysadmind, line 54: I think it helps readability to put the lock/mutex just before the resources it is protecting and add a blank line before the lock and a blank line after the protected resources just to visually see exactly what the lock is for.

collector/snapshots.go
```diff
@@ -244,6 +255,15 @@ func (s *Snapshots) getAndParseURL(u *url.URL, data interface{}) error {
 }
 
 func (s *Snapshots) fetchAndDecodeSnapshotsStats() (map[string]SnapshotStatsResponse, error) {
+	resp, err, shared := s.group.Do("snapshots", func() (interface{}, error) {
+		_ = level.Debug(s.logger).Log("msg", "getting snapshots metrics")
```
sysadmind, line 259: `fetching snapshots metrics` might be slightly more clear

sysadmind (approved):
Just a few nitpicks. I like that the default is still the original behavior. It would be cool to add a test to verify the behavior but probably not trivial to implement.

ranjithruban:
@caarlos0 @sysadmind is it possible to add this in new version?. thank you

PR #342, opened 2020-04-27
add jvm.mem.heap_used_percent

lechen26:
we needed the percentage JVM memory used by Heap in order to monitor when we reach 85% utilization of memory, same as we do for cpu.

output in our case looks like the following:
```
# HELP elasticsearch_jvm_memory_used_percent percent of JVM memory used
# TYPE elasticsearch_jvm_memory_used_percent gauge
elasticsearch_jvm_memory_used_percent{area="heap",cluster="xxxxxxx",es_client_node="true",es_data_node="true",es_ingest_node="true",es_master_node="true",host="xxx",name="instance-1"} 79
elasticsearch_jvm_memory_used_percent{area="heap",cluster="xxxxxx",es_client_node="true",es_data_node="true",es_ingest_node="true",es_master_node="true",host="xxxxx",name="instance-2"} 69
elasticsearch_jvm_memory_used_percent{area="heap",cluster="xxxxxx",es_client_node="true",es_data_node="true",es_ingest_node="true",es_master_node="true",host="xxxxx",name="instance-3"} 73
```

### DIFF
--- a/collector/nodes.go
+++ b/collector/nodes.go
@@ -1180,6 +1180,20 @@ func NewNodes(logger log.Logger, client *http.Client, url *url.URL, all bool, no
 					return append(defaultNodeLabelValues(cluster, node), "young")
 				},
 			},
+                        {
+                                Type: prometheus.GaugeValue,
+                                Desc: prometheus.NewDesc(
+                                        prometheus.BuildFQName(namespace, "jvm_memory", "used_percent"),
+                                        "percent of JVM memory used",
+                                        append(defaultNodeLabels, "area"), nil,
+                                ),
+                                Value: func(node NodeStatsNodeResponse) float64 {
+                                        return float64(node.JVM.Mem.HeapPercent)
+                                },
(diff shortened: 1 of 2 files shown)


frenkye:
IMO there is no need for this addition. You can do alert like this:
```
round(elasticsearch_jvm_memory_used_bytes{area="heap"} / elasticsearch_jvm_memory_max_bytes * 100, 0.1) > 85
```

lechen26:
@frenkye that's right.
but we use to  export the metrics from the   elastic exporter  to stackdriver and there we set alerts, not using the alert manager.

but  10x

mattwelke:
@lechen26 That's our situation too. Maybe the new MQL approach (https://cloud.google.com/monitoring/mql) in Monitoring (previously known as Stackdriver) would allow us to set alerts based on just these exported metrics?

SuperQ:
This is unnecessary as @frenkye says. If you need a derived metric, you can use a recording rule.

PR #351, opened 2020-05-25
add new options to specify list of exported indices

msvechla:
Hi, 

First of all thanks for this awesome project!

As the `/_all/_stats` and `/_all/_settings` requests can be resource extensive, this PR adds options to specify a comma separated list of indices or wildcard index expressions, based on which the list of to be exported indices is determined.

This works in the same way as the `<index>` parameter here: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-stats.html

Options Added:

```
es.indices-list (default: *)
es.indices_settings-list (default: *)
```

Let me know if this is fine, or if there is anything else that needs to be adjusted.
Thanks for your help!


ebernhardson:
I can't speak to the particulars of the implementation done here, but as a feature this is something I think would be useful to include. We have thousands of indices across multiple clusters, but our use case only needs index specific metrics for a few of the largest indices.

PR #364, opened 2020-07-08
#257: Add gauge for total fields mapped in an index

### LINKED ISSUE #257: Add gauge for total fields mapped in an index
Motivation: ES imposes a limit on the number of fields mapped, and we'd like to alert when we're approaching that limit. We've got a lot of custom ELK tooling and suchlike, so there's the potential for a bug to flood the index with new fields, for example. 

Happy to bring a PR if it is thought to be useful!

msodi:
closes #257

### DIFF
--- a/collector/indices_mappings.go
+++ b/collector/indices_mappings.go
@@ -1,0 +1,182 @@
+package collector
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"path"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	defaultIndicesMappingsLabels = []string{"index"}
+)
+
+type indicesMappingsMetric struct {
+	Type  prometheus.ValueType
+	Desc  *prometheus.Desc
+	Value func(indexMapping IndexMapping) float64
+}
+
+// IndicesMappings information struct
+type IndicesMappings struct {
+	logger log.Logger
+	client *http.Client
+	url    *url.URL
+
+	up                              prometheus.Gauge
+	totalScrapes, jsonParseFailures prometheus.Counter
+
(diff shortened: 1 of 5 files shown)


collector/indices_mappings.go
```diff
@@ -0,0 +1,182 @@
...
+				),
+				Value: func(indexMapping IndexMapping) float64 {
+					return countFieldsRecursive(indexMapping.Mappings.Properties, 0)
+				},
+			},
+		},
+	}
+}
+
+func countFieldsRecursive(properties Properties, fieldCounter float64) float64 {
```
sysadmind, line 74: This probably makes sense as a pointer to reduce the number of memory allocations for each recursion depth. That would also mean that there is no return necessary.

collector/indices_mappings_response.go
```diff
@@ -0,0 +1,34 @@
...
+// IndicesMappingsResponse is a representation of elasticsearch mappings for each index
+type IndicesMappingsResponse map[string]IndexMapping
+
+// IndexMapping defines the struct of the tree for the mappings of each index
+type IndexMapping struct {
+	Mappings Mappings `json:"mappings"`
+}
+
+// Mappings defines all index mappings
+type Mappings struct {
```
sysadmind, line 12: I think the names `Mappings`, `Properties`, `Fields`, `Property`, and `Field` are really generic and could cause naming collisions in the future. It probably makes sense to use more specific names like `IndexMappings`, `IndexMappingField`, etc.

jnadler:
@msodi are you still there?   we need this as well, and i'd be willing to work on this PR if that helps

msodi:
> @msodi are you still there? we need this as well, and i'd be willing to work on this PR if that helps

I am still there but did not have the chance yet to continue working on this PR. Feel free to do so. Thanks @jnadler

jnadler:
Please close in favor of #386 and I'll try to push this over the finish line.

msodi:
Thanks @jnadler for taking over

PR #388, opened 2020-09-18
Update README.md

moritzluedtke:
- Updated the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,11 +21,10 @@ Example `docker-compose.yml`:
 ```yaml
 elasticsearch_exporter:
     image: justwatch/elasticsearch_exporter:1.1.0
-    command:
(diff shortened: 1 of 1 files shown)


moritzluedtke:
I just started using the exporter. It works great so thanks for your work first of all!

The example docker-compose file had an error in IntelliJ and I found some parts of the README difficult to understand so I wrote some sentences which would have made it a lot clearer and faster for me to get it running the way I want it.

I'm curious for your input and feedback.

sysadmind:
This PR is out of date. I am closing but if you make updates to the source, you can reopen. The docker-compose in the README looks ok to me. I do not thing adding the second example makes the intent any more clear, but adding some text may be useful. The environment variables have also been removed for the most part.

PR #396, opened 2020-11-25
feat: allow Prometheus namespace to be customized

rdumont:
I need to use a different Prometheus namespace, so I made it configurable with `prom.namespace` or `PROM_NAMESPACE`.


SuperQ:
This doesn't seem like it's useful to the general userbase. If you have specific restrictions, I recommend using `metric_relabel_configs`.

rdumont:
That makes sense, @SuperQ! Thanks.

PR #424, opened 2021-06-01
feat: add duration metrics for collectors

sysadmind:
adds metrics to capture duration of time spent collecting metrics for:
  - cluster_health
  - clustersettings
  - index_stats
  - node_stats
  - snapshot_stats

this feature was adapted from the caarlos0/elasticsearch_exporter repo.

There seem to be a lot of metrics missing from the README, so I just added the metrics that this PR addresses, but I want to ensure that all metrics are in the README in a future PR.

### DIFF
--- a/collector/cluster_health.go
+++ b/collector/cluster_health.go
@@ -56,7 +56,7 @@ type ClusterHealth struct {
 
 	up                prometheus.Gauge
 	totalScrapes      prometheus.Counter
-	totalScrapeTime   prometheus.Counter
+	scrapeDuration    prometheus.Gauge
 	jsonParseFailures prometheus.Counter
 
 	metrics      []*clusterHealthMetric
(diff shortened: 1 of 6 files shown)


SuperQ:
Hmm, typically we add this kind of thing as a gauge. For example in the node_exporter, [this is how we handle it](https://github.com/prometheus/node_exporter/blob/5b13775dc868e6c30a2346a8885fcf067c7aad57/collector/collector.go#L176).

collector/snapshots.go
```diff
@@ -78,6 +81,10 @@ func NewSnapshots(logger log.Logger, client *http.Client, url *url.URL) *Snapsho
 			Name: prometheus.BuildFQName(namespace, "snapshot_stats", "total_scrapes"),
 			Help: "Current total ElasticSearch snapshots scrapes.",
 		}),
+		scrapeDuration: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: prometheus.BuildFQName(namespace, "snapshot_stats", "scrape_duration_seconds"),
+			Help: "Duration spent in ElasticSearch snapshots scrape.",
```
caarlos0, line 86: Maybe "Time spent" sounds better? 🤔

caarlos0 (approved):

collector/snapshots.go
SuperQ: I like "Time spent" as well.

collector/cluster_health.go
```diff
@@ -77,6 +80,10 @@ func NewClusterHealth(logger log.Logger, client *http.Client, url *url.URL) *Clu
 			Name: prometheus.BuildFQName(namespace, subsystem, "total_scrapes"),
 			Help: "Current total ElasticSearch cluster health scrapes.",
 		}),
+		scrapeDuration: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: prometheus.BuildFQName(namespace, subsystem, "scrape_duration_seconds"),
```
SuperQ, line 84: Hmm, maybe we should flip these per-subsystem metrics around and make them global, and use a label.

Maybe we do this in a refactor before cutting a new release? Although, we should not merge this until we cut a "safe harbor" release.

SuperQ (approved):
LGTM. I like the metric, but the whole per-collector metric naming stuff needs refactoring.

LaurentDumont:
Curious in this one too. I'd like to see if we have a performance issues with queries and it's a bit hard to see right now.

SuperQ (changes requested):
Ping, this has some comments to address.

SuperQ:
This needs a rebase.

smowky:
this metrics might be really useful, any chance this PR will get processed?

PR #426, opened 2021-06-02
fix(test): resolve test race conditions

sysadmind:
Fix a race condition in pkg/clusterinfo mockConsumer by protecting the data with a mutex. This resolves the issue where `make` fails to complete the build and tests.

### DIFF
--- a/pkg/clusterinfo/clusterinfo_test.go
+++ b/pkg/clusterinfo/clusterinfo_test.go
@@ -21,6 +21,7 @@ import (
 	"net/url"
 	"os"
 	"reflect"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/clusterinfo/clusterinfo_test.go
```diff
@@ -85,7 +89,9 @@ func newMockConsumer(ctx context.Context, name string, t *testing.T) *mockConsum
 		for {
 			select {
 			case d := <-mc.ch:
+				mc.mu.Lock()
 				mc.data = d
+				mc.mu.Unlock()
 				t.Logf("consumer %s received data from channel: %+v\n", mc, mc.data)
```
roidelapluie, line 95: there should still be a lock here. I would not take a read lock but extend the write log.

roidelapluie (commented):
Awesome! What about we also change the .circleci in this PR to validate?

pkg/clusterinfo/clusterinfo_test.go
sysadmind: Thank you for the feedback! Nice catch. This should now be resolved.

sysadmind:
I don't have experience with circleci, but I think I have adjusted the config file appropriately. I referenced another exporter so if it is not the correct change, please let me know.

roidelapluie (approved):

roidelapluie:
Thanks!

Issue #419, opened 2021-05-27
Elasticsearch 7.13 Breaking Metrics


mcmarkj:
Elasticsearch 7.13 appears to have removed several key metrics around cluster disk storage etc. 

This means metrics like `elasticsearch_filesystem_data_size_bytes` are no longer present.

epicvinny:
Same here, error log:

caller=nodes.go:1851 msg="failed to fetch and decode node stats" err="json: cannot unmarshal array into Go struct field NodeStatsNodeResponse.http of type int"

jivco:
I think the problem is that they have [added info on each HTTP client to HTTP stats](https://github.com/elastic/elasticsearch/pull/64561) and this breaks the nodes_response collector.

adammike:
Is anyone actively looking at this issue?

hansh0801:
I also have this problem. Is anyone actively looking at this issue?

egallis31:
> I think the problem is that they have [added info on each HTTP client to HTTP stats](https://github.com/elastic/elasticsearch/pull/64561) and this breaks the nodes_response collector.
> 
> I think this is a fast solution but I don't know how to export the new metrics. I'm using [this branch](https://github.com/repl-mike-roest/elasticsearch_exporter/tree/fix-build-go-mod) for building because the original one is broken.
> 
> [diff.txt](https://github.com/prometheus-community/elasticsearch_exporter/files/6570460/diff.txt)

Getting some issues building using that branch - anything missing from the make file?


```
>> building docker image
Sending build context to Docker daemon  11.51MB
Step 1/9 : FROM quay.io/prometheus/golang-builder as builder
 ---> a6c160c7ad69
Step 2/9 : ADD .   /go/src/github.com/justwatchcom/elasticsearch_exporter
 ---> Using cache
 ---> 34bc37901c23
Step 3/9 : WORKDIR /go/src/github.com/justwatchcom/elasticsearch_exporter
 ---> Using cache
 ---> 2e4434305a5e
Step 4/9 : RUN make
 ---> Running in 2ecf04b34fff
>> formatting code
go: downloading github.com/prometheus/promu v0.12.0
go: downloading github.com/google/go-github/v25 v25.1.3
go: downloading github.com/pkg/errors v0.9.1
go: downloading github.com/prometheus/common v0.25.0
go: downloading go.uber.org/atomic v1.7.0
[...]

PR #427, opened 2021-06-03
test: refactor TestNodesStats

sysadmind:
As per #419, this functionality is failing for elasticsearch version 7.13+ due to a breaking change in the response from elasticsearch. This adds some test fixtures for a range of elasticsearch versions including 7.13 to help ensure that future changes fix the incompatibility with 7.13 and maintain compatibility with versions < 7.13.

To reviewers: It may be easier to view with whitespace turned off due to the control structures adjusting indentation. 

**NOTE: This does not include an actual fix for the 7.13 response, only the failing test.**


SuperQ (approved):
LGTM, thanks!

caarlos0 (approved):

PR #436, opened 2021-06-11
Update deployment.yml

oldthreefeng:
bump elasticsearch-exporter to v1.2.0.

### DIFF
--- a/examples/kubernetes/deployment.yml
+++ b/examples/kubernetes/deployment.yml
@@ -23,7 +23,7 @@ spec:
         - /bin/elasticsearch_exporter
         - --es.uri=http://elasticsearch:9200
         - --es.all
-        image: quay.io/prometheuscommunity/elasticsearch-exporter:1.1.0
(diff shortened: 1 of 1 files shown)


examples/kubernetes/deployment.yml
```diff
@@ -23,7 +23,7 @@ spec:
         - /bin/elasticsearch_exporter
         - --es.uri=http://elasticsearch:9200
         - --es.all
-        image: quay.io/prometheuscommunity/elasticsearch-exporter:1.1.0
```
sysadmind, line 26: This should be `latest` so that we do not need to update this example on every release.

sysadmind (changes requested):
Thanks for the update. The CI check is failing for the DCO.

examples/kubernetes/deployment.yml
oldthreefeng: done

sysadmind (approved):

Issue #415, opened 2021-05-17
failed to fetch and decode snapshot stats" err:"json: cannot unmarshal object into Go struct field .settings of type string"


bydeath:
it failed to scrape snapshots stats

warn log:
snapshots.go:265 msg "failed to fetch and decode snapshot stats" err:"json: cannot unmarshal object into Go struct field .settings of type string"

I am using the exporter of latest version V1.1.0, elasticsearch version is 7.9.2.

bydeath:
To reproduce the issue, install the elasticsearch plugin repository-hdfs https://www.elastic.co/guide/en/elasticsearch/plugins/current/repository-hdfs-config.html#repository-hdfs-config
``` and create a hdfs type repository like this. To reproduce, `conf` field should be set. 
PUT _snapshot/my_hdfs_repository
{
  "type": "hdfs",
  "settings": {
    "uri": "hdfs://namenode:8020/",
    "path": "elasticsearch/repositories/my_hdfs_repository",
    "conf.dfs.client.read.shortcircuit": "true"
  }
}
```

JohnMops:
Any news on this one ?

sysadmind:
#442 addressed this issue.

autokilla47:
+1

PR #442, opened 2021-06-17
Fix snapshot_stats metrics which happens on HA hdfs type snapshot rep…

bydeath:
…ository

fix issue https://github.com/prometheus-community/elasticsearch_exporter/issues/415
err logs: "failed to fetch and decode snapshot stats" err:"json: cannot unmarshal object into Go struct field .settings of type string" "

repository-hdfs is an elasticsearch plugin: https://www.elastic.co/guide/en/elasticsearch/plugins/current/repository-hdfs-config.html#repository-hdfs-config

### DIFF
--- a/collector/snapshots_reponse.go
+++ b/collector/snapshots_reponse.go
@@ -43,6 +43,6 @@ type SnapshotStatDataResponse struct {
 
 // SnapshotRepositoriesResponse is a representation snapshots repositories
 type SnapshotRepositoriesResponse map[string]struct {
-	Type     string            `json:"type"`
-	Settings map[string]string `json:"settings"`
+	Type     string                 `json:"type"`
+	Settings map[string]interface{} `json:"settings"`
 }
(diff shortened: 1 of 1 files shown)


collector/snapshots_reponse.go
```diff
@@ -43,6 +43,6 @@ type SnapshotStatDataResponse struct {
 
 // SnapshotRepositoriesResponse is a representation snapshots repositories
 type SnapshotRepositoriesResponse map[string]struct {
-	Type     string            `json:"type"`
-	Settings map[string]string `json:"settings"`
+	Type     string                 `json:"type"`
+	Settings map[string]interface{} `json:"settings"`
```
caarlos0, line 47: if we don't use the value of settings anywhere, we could just remove the field 🤔

bydeath: > if we don't use the value of settings anywhere, we could just remove the field `Settings`
It seems the value of settings is not used now. I tried it  worked without `Settings` field, and it's another way to fix the bug.

sysadmind (approved):

PR #353, opened 2020-05-26
Aws es signing

JackFazackerley:
Adds AWS ElasticSearch request signing, closes #346,  also merged #333 into my branch as I wanted Go modules and it was already created as a PR.

PR #443, opened 2021-06-17
Add support for Amazon Elasticsearch Service

JamiKarvanen:
This is based on @JackFazackerley's initial work from here: https://github.com/prometheus-community/elasticsearch_exporter/pull/353

I rebased this against the latest master and added additional credential provider, so this also works with pod IAM roles when running in EKS.


main.go
```diff
@@ -85,6 +93,10 @@ func main() {
 		esInsecureSkipVerify = kingpin.Flag("es.ssl-skip-verify",
 			"Skip SSL verification when connecting to Elasticsearch.").
 			Default("false").Envar("ES_SSL_SKIP_VERIFY").Bool()
+		esAWS = kingpin.Flag("es.aws", "Enable support for Amazon Elasticsearch Service").
+			Default("false").Envar("ES_AWS").Bool()
+		esAWSRegion = kingpin.Flag("es.aws-region", "Sets the AWS region").
+				Default("").Envar("ES_AWS_REGION").String()
```
JamiKarvanen, line 99: I have no idea why `go fmt` enforces this indentation... Would fail the CircleCI tests otherwise

roidelapluie:
Can we leverage https://github.com/prometheus/common/tree/main/sigv4 for the sigv4 part? does it make sense?

JamiKarvanen:
I think we can. Great idea! I'll refactor this to use it instead

justinpjose:
Any progress with this ticket? It would be very useful for AWS ES signing support to be included for us as we require it

visla-xugeng:
I am using AWS elasticsearch. I plan to use chart, prometheus-elasticsearch-exporter, to monitor it. A quick question here:
what's the uri of es? Is the Domain endpoint (VPC) uri? 

es:
  uri: http://localhost:9200

PR #456, opened 2021-07-16
Grafana dashboard: remove the Example from the name, allowing direct and easy provisionning use.

desaintmartin:


sysadmind (approved):
LGTM Thanks

Issue #359, opened 2020-06-24
[PROPOSAL] - develop monitoring mixins


P6rguVyrst:
I'd be happy to volunteer for this, if this kind of contribution would be welcome.

Examples:
https://github.com/thanos-io/thanos/tree/master/mixin
https://github.com/kubernetes-monitoring/kubernetes-mixin
https://github.com/prometheus/node_exporter/tree/master/docs/node-mixin

There's an existing mixin for Elasticsearch, but it uses metrics from a different exporter: https://github.com/lukas-vlcek/elasticsearch-mixin

I believe the mixin should be where metrics are defined, for this exporter - here.

Design document for the concept: https://docs.google.com/document/d/1A9xvzwqnFVSOZ5fD3blKODXfsat5fg6ZhnKu9LK3lB4/edit
Presentation for the concept: https://www.youtube.com/watch?v=GDdnL5R_l-Y

P6rguVyrst:
Is the project dead? 
Will the PR be accepted if I decide to invest any time in this?

P6rguVyrst:
Hmm, it looks like this project has changed owners? 
Are PRs to add mixins welcome now?

metalmatze:
Yes, we should totally add a mixin for sure!

P6rguVyrst:
Sorry, I just noticed the feedback. 

I have something defined on [this fork](https://github.com/transferwise/elasticsearch_exporter/tree/master/mixin), but would want to have another internal review of this before raising a PR here.

PR #458, opened 2021-07-21
[ISSUE-359] Monitoring mixins

P6rguVyrst:
Resolves: https://github.com/prometheus-community/elasticsearch_exporter/issues/359

Alerts and Rules are something I've created for our needs. Dashboard is something I borrowed from grafana net and modified... Dashboard also has cert expiry which could be a bit too specific for our setup..

Haven't had time/willingness to properly define the dashboard, but this could be better than nothing?

es-dashboard

My main issue with the grafana net dashboard was that it had hardcoded datasource and I think it didn't work well if there was more than 1 exporter gathering data for all the nodes in the cluster.

In our setup we extend on this even more with **ElasticsearchIndexingStopped** and **ElasticsearchClusterDegraded** alerts, but these are too specific for our setup and needs, and have thus been kept separate from the exporter code.

Sorry for all the excess PRs maybe they can be squashed or I can raise a new PR with git diff from upstram..


P6rguVyrst:
Hmm.. Following the DCO check instructions created much more change than I wanted :/

PR #459, opened 2021-07-24
Add the ability to use ApiKey based authentication

### LINKED ISSUE #454: Not able to use API Keys from Elastic Cloud
We're running Elasticsearch on Elastic Cloud and they don't allow Basic Auth as authentication method.
You can only use [API Keys](https://www.elastic.co/guide/en/cloud/current/ec-api-authentication.html).

Unfortunately I can't find a way to use this Elasticsearch exporter with API Keys. I tried to include the id and key as username and password in the URI in different ways, but Elasticsearch doesn't accept that.

Is there a way to use an API key as authentication method?

TrafeX:
If you're using Elasticsearch on Elastic Cloud the only way to authenticate is using [API Keys](https://www.elastic.co/guide/en/cloud/current/ec-api-authentication.html). 
This PR adds the ability to define the API Key and set it as Authorization header with every request.

Please let me know if I can improve anything about the implementation!

Fixes #454

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ elasticsearch_exporter --help
 | es.client-cert          | 1.0.2                 | Path to PEM file that contains the corresponding cert for the private key to connect to Elasticsearch. | |
 | es.clusterinfo.interval | 1.1.0rc1              |  Cluster info update interval for the cluster label | 5m |
 | es.ssl-skip-verify      | 1.0.4rc1              | Skip SSL verification when connecting to Elasticsearch. | false |
+| es.apiKey               | unreleased            | Skip SSL verification when connecting to Elasticsearch. | false |
 | web.listen-address      | 1.0.2                 | Address to listen on for web interface and telemetry. | :9114 |
 | web.telemetry-path      | 1.0.2                 | Path under which to expose metrics. | /metrics |
 | version                 | 1.0.2                 | Show version info on stdout and exit. | |
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ elasticsearch_exporter --help
 | es.client-cert          | 1.0.2                 | Path to PEM file that contains the corresponding cert for the private key to connect to Elasticsearch. | |
 | es.clusterinfo.interval | 1.1.0rc1              |  Cluster info update interval for the cluster label | 5m |
 | es.ssl-skip-verify      | 1.0.4rc1              | Skip SSL verification when connecting to Elasticsearch. | false |
-| es.apiKey               | unreleased            | Skip SSL verification when connecting to Elasticsearch. | false |
+| es.apiKey               | unreleased            | API Key to use for authenticating against Elasticsearch. |  |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -56,6 +56,7 @@ elasticsearch_exporter --help
 | es.client-cert          | 1.0.2                 | Path to PEM file that contains the corresponding cert for the private key to connect to Elasticsearch. | |
 | es.clusterinfo.interval | 1.1.0rc1              |  Cluster info update interval for the cluster label | 5m |
 | es.ssl-skip-verify      | 1.0.4rc1              | Skip SSL verification when connecting to Elasticsearch. | false |
+| es.apiKey               | unreleased            | Skip SSL verification when connecting to Elasticsearch. | false |
```
sysadmind, line 59: This comment is wrong and so is the default.

main.go
```diff
@@ -31,6 +31,16 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
+type transportWithApiKey struct {
+	underlyingTransport http.RoundTripper
+	apiKey              *string
```
sysadmind, line 36: I don't think that this needs to be a pointer. It won't change after start.

main.go
```diff
@@ -31,6 +31,16 @@ import (
...
+}
+
+func (t *transportWithApiKey) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Add("Authorization", "ApiKey "+*t.apiKey)
```
sysadmind, line 40: I think `fmt.Sprintf()` would make this more readable instead of string concatenation.

main.go
```diff
@@ -114,12 +127,24 @@ func main() {
...
-		Timeout: *esTimeout,
-		Transport: &http.Transport{
-			TLSClientConfig: tlsConfig,
-			Proxy:           http.ProxyFromEnvironment,
-		},
+		Timeout:   *esTimeout,
+		Transport: httpTransport,
+	}
+
+	if *esApiKey != "" {
```
sysadmind, line 140: I think it would make sense to move this if statement above the httpClient variable above. That would allow only overwriting the httpTransport variable inside the if statement. That would mean there is only one place where the httpClient variable is set.

sysadmind (changes requested):
Thanks for the PR. Just a few changes that I would prefer to see to clean it up.

main.go
TrafeX: Agreed, changed!

main.go
TrafeX: I agree and I tried to only change the `Transport`, but the issue is that both transports don't have the same type. So I couldn't figure it out how to use one variable and assign either one of the transports based on the `esApiKey` variable.
Do you know how to do this?

sysadmind: You can define the var ahead of time as the interface type (i.e. `var transport http.RoundTripper`)

TrafeX: Ah of course, I needed to use the interface. It's changed, thanks.

sysadmind (approved):
LGMT

PR #468, opened 2021-08-20
node stats: introduce pipeline metrics

preludedrew:
* Will look into introducing processor metrics at a later time


nsaboren:
@SuperQ @sysadmind I apologize for the ping. I'm curious if I need an issue associated to this, or possibly a sponsor?

preludedrew:
@SuperQ @sysadmind I apologize for the ping. I'm curious if I need an issue associated to this, or possibly a sponsor?

collector/nodes.go
```diff
@@ -88,6 +88,7 @@ var (
 	defaultRoleLabels               = []string{"cluster", "host", "name"}
 	defaultThreadPoolLabels         = append(defaultNodeLabels, "type")
 	defaultBreakerLabels            = append(defaultNodeLabels, "breaker")
+	defaultPipelineLabels           = append(defaultNodeLabels, "pipeline")
```
sysadmind, line 91: I don't think that adding another default label is a good idea here. Labels increase cardinality on metrics and that can tax prometheus.

sysadmind (changes requested):

collector/nodes.go
nsaboren: Understood about adding labels and cardinality, but I'm not sure there's a better way around it. Multiple pipelines can be configured, all with different names. We'll need to ability to aggregate or filter based on the name. The only other method I can think of is to put the pipeline name in the metric name and extract it there, but that's just passing the buck.

preludedrew: Understood about adding labels and cardinality, but I'm not sure there's a better way around it. Multiple pipelines can be configured, all with different names. We'll need to ability to aggregate or filter based on the name. The only other method I can think of is to put the pipeline name in the metric name and extract it there, but that's just passing the buck.

sysadmind: I'm not saying that these metrics should not have those labels, just that these labels don't need to be default labels for all node metrics. I think over time this list of default node labels will actually go away. 

What I suggest is that instead of just using defaultNodeLabels on the metric labels, only define the 2 or 3 that are necessary which seem to be cluster name, node, and pipeline. Does that make sense? The other labels aren't used on the metric anyway so there's no need to define them on the metric.

PR #498, opened 2021-11-10
Remove non-existing CPU metrics

multani:
I couldn't find any Elasticsearch versions where these CPU metrics
were exposed, so they are always set to 0.


SuperQ:
I can't confirm all versions, but I can say that these are always 0 for my 7.x setup

SuperQ (approved):

PR #518, opened 2022-01-19
Cleanup non-standard ENV var setup

SuperQ:
Drop ENV vars from flags as this is non-standard in the Prometheus
ecosystem.

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/main.go
+++ b/main.go
@@ -48,67 +48,64 @@ func main() {
 	var (
 		listenAddress = kingpin.Flag("web.listen-address",
 			"Address to listen on for web interface and telemetry.").
-			Default(":9114").Envar("WEB_LISTEN_ADDRESS").String()
+			Default(":9114").String()
 		metricsPath = kingpin.Flag("web.telemetry-path",
 			"Path under which to expose metrics.").
-			Default("/metrics").Envar("WEB_TELEMETRY_PATH").String()
+			Default("/metrics").String()
 		esURI = kingpin.Flag("es.uri",
 			"HTTP API address of an Elasticsearch node.").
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -48,67 +48,67 @@ func main() {
...
 			"Path to PEM file that contains the corresponding cert for the private key to connect to Elasticsearch.").
-			Default("").Envar("ES_CLIENT_CERT").String()
+			Default("").String()
 		esInsecureSkipVerify = kingpin.Flag("es.ssl-skip-verify",
 			"Skip SSL verification when connecting to Elasticsearch.").
-			Default("false").Envar("ES_SSL_SKIP_VERIFY").Bool()
+			Default("false").Bool()
 		esApiKey = kingpin.Flag("es.apiKey",
 			"API Key to use for authenticating against Elasticsearch").
-			Default("").Envar("ES_API_KEY").String()
```
sysadmind, line 102: I think this one should stay because it's sensitive.

sysadmind (changes requested):

main.go
SuperQ: Good point, fixed.

SuperQ: Actually, we should probably remove this flag.

sysadmind: This was added recently: https://github.com/prometheus-community/elasticsearch_exporter/pull/459
It supports elastic cloud instances that use the API key for auth. Are you suggesting an alternative or that it is not necessary at all?

sysadmind (approved):

sysadmind (approved):

PR #520, opened 2022-01-19
Update process CPU metric

SuperQ:
With PR #498, we only have one process CPU metric.
* Drop the label.
* Update the name to match Prometheus naming.

Signed-off-by: SuperQ <superq@gmail.com>


SuperQ:
CC @multani, followup to your change.

multani:
Thanks for the heads up @SuperQ, looks good to me :+1:

sysadmind (approved):
LGTM

PR #521, opened 2022-01-19
Add a common collector interface

sysadmind:
This provides the groundwork to refactor the collectors into a more consistent interface. It borrows some influence from the node_exporter with the goal to easily filter in/out collector modules. The execute func also provides a context for potential cancellation.

I'm open to feedback to shape this up before more serious refactors.

### DIFF
--- a/collector/cluster_health.go
+++ b/collector/cluster_health.go
@@ -26,10 +26,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-const (
-	namespace = "elasticsearch"
-)
-
 var (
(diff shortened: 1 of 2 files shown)


SuperQ (approved):
LGTM, I really would like to fully abstract this collector registration into the exporter-toolkit, but I need to think about how to design things such that we can pass optional config like esURL.

For now, this is great. If we ever build a generic collector registration package, it should make this easy to convert.

sysadmind:
@SuperQ Thanks for the review. I agree that it would be nice to have something consistent. Maybe this can serve as a project to prove out an interface/implementation that can be adopted to the exporter-toolkit.

Issue #330, opened 2020-04-03
I'm not familiar with the ES exporter data. There's a gap problem


guleng:
I'm not familiar with the ES exporter data. There's a gap problem

My profile：
`        - --es.uri=http://elasticsearch-logging.kube-system:9200 
        - --es.all
        - --es.indices
        - --es.timeout=30s
        - --es.clusterinfo.interval=30s
        - --es.shards
        - --es.snapshots
        - --web.listen-address=:9108
        - --web.telemetry-path=/metrics`

![image](https://user-images.githubusercontent.com/40238640/78328708-62c61a80-75b2-11ea-89fc-366cddf90b33.png)

I don't know why.

Thakurvaibhav:
Hi @guleng , 
I am running into the same issue. Were you able to resolve it ?

wbira:
Hello,

I have same issue, any news on this? Or any advise how to troubleshoot this? Exporter logs looking fine. No issues with stability of elastic cluster

`level=info ts=2021-12-08T23:56:45.262624829Z caller=clusterinfo.go:169 msg="providing consumers with updated cluster info label"
level=info ts=2021-12-09T00:01:45.262650571Z caller=clusterinfo.go:169 msg="providing consumers with updated cluster info label"
level=info ts=2021-12-09T00:06:45.262672569Z caller=clusterinfo.go:169 msg="providing consumers with updated cluster info label"
level=info ts=2021-12-09T00:11:45.262654392Z caller=clusterinfo.go:169 msg="providing consumers with updated cluster info label"
level=info ts=2021-12-09T00:16:45.262670029Z caller=clusterinfo.go:169 msg="providing consumers with updated cluster info label"
level=info ts=2021-12-09T00:21:45.262674688Z caller=clusterinfo.go:169 msg="providing consumers with updated cluster info label"
level=info ts=2021-12-09T00:26:45.262659058Z caller=clusterinfo.go:169 msg="providing consumers with updated cluster info label"
level=info ts=2021-12-09T00:31:45.262651624Z caller=clusterinfo.go:169 msg="providing consumers with updated cluster info label"
level=info ts=2021-12-09T00:36:45.262712409Z caller=clusterinfo.go:169 msg="providing consumers with updated cluster info label"
level=info ts=2021-12-09T00:41:45.26268414Z caller=clusterinfo.go:169 msg="providing consumers with updated cluster info label"
[...]

PR #529, opened 2022-02-04
add timeout to node stats

trk-bxn:
- Added fix for metric gabs on high load on single node.

---
With the elasticsearch version 7 and Opensearch version 1 there is a new timeout parameter implement.

Elasticsearch 6 ignores this timeout parameter and this change have no impact.

Tested with elasticsearch 6.8 and opensearch 1.2

Elasticsearch
https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-stats.html#cluster-nodes-stats-api-query-params

Opensearch
https://opendistro.github.io/for-elasticsearch-docs/docs/elasticsearch/rest-api-reference/#nodesinfo


sysadmind (commented):
It looks like the httpclient passed into collector.NewNodes already has this set on the client side. I think you could pull it from the http.Client.Timeout field so the signature does not need to change. I am also not sure what the benefit is of the http query string parameter. Can you explain how this would benefit users of this exporter?

trk-bxn:
this fix this problem in newer elasticsearch version
https://github.com/prometheus-community/elasticsearch_exporter/issues/330

PR #536, opened 2022-02-23
Refactor cluster info collector

sysadmind:
Adds new cluster info collector. This will eventually replace the cluster info collector in pkg/clusterinfo. The current limitation in fully replacing that package is that it provides downstream collectors with the cluster info to apply the cluster name label. I plan to find a better solution for this when I refactor the next collector.

Updates the Collector interactions to more closely match node_exporter and postgres_exporter style of registering collectors.


SuperQ (approved):
LGTM

Issue #555, opened 2022-03-24
Expose SLM Metrics


Evesy:
The exporter currently exposes metrics directly around snapshots but it would be good to also expose metrics from SLM [1] around scheduled snapshots.

When using SLM snapshots a scheduled snapshot can fail without the underlying Elastic snapshots having any failure, e.g. with `ignore_unavailable` set to false if I purposefully turn a index red you can get a snapshot to fail like the below:

```
{
  "type": "snapshot_exception",
  "reason": "[gcs:kibana-2022.03.24-rmug_mrwtauw_hac6mjchg/uw4vqewLQVSTUDMWUxZutA] Indices don't have primary shards [.kibana_7.15.1_001, .kibana-event-log-7.15.1-000002, .kibana_task_manager_pre7.4.0_001, .kibana_task_manager_7.15.1_001]",
[...]

PR #558, opened 2022-04-01
Add additional collector for SLM stats

Evesy:
Closes https://github.com/prometheus-community/elasticsearch_exporter/issues/555

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,20 +224,20 @@ Further Information
 | elasticsearch_clusterinfo_up                                          | gauge     | 1           | Up metric for the cluster info collector
 | elasticsearch_clusterinfo_version_info                                | gauge     | 6           | Constant metric with ES version information as labels
 | elasticsearch_slm_stats_up                                            | gauge     | 0           | Up metric for SLM collector
-| elasticsearch_slm_stats_total_scrapes                                 | gauge     | 0           | Number of scrapes for SLM collector
-| elasticsearch_slm_stats_json_parse_failures                           | gauge     | 0           | JSON parse failures for SLM collector
-| elasticsearch_slm_stats_retention_runs                                | gauge     | 0           | Total retention runs
-| elasticsearch_slm_stats_retention_failed                              | gauge     | 0           | Total failed retention runs
-| elasticsearch_slm_stats_retention_timed_out                           | gauge     | 0           | Total retention run timeouts
-| elasticsearch_slm_stats_retention_retention_deletion_time_millis      | gauge     | 0           | Retention run deletion time
-| elasticsearch_slm_stats_total_snapshots_taken                         | gauge     | 0           | Total snapshots taken
-| elasticsearch_slm_stats_total_snapshots_failed                        | gauge     | 0           | Total snapshots failed
-| elasticsearch_slm_stats_total_snapshots_deleted                       | gauge     | 0           | Total snapshots deleted
-| elasticsearch_slm_stats_total_snapshots_failed                        | gauge     | 0           | Total snapshots failed
-| elasticsearch_slm_stats_snapshots_taken                               | gauge     | 1           | Snapshots taken by policy
-| elasticsearch_slm_stats_snapshots_failed                              | gauge     | 1           | Snapshots failed by policy
-| elasticsearch_slm_stats_snapshots_deleted                             | gauge     | 1           | Snapshots deleted by policy
-| elasticsearch_slm_stats_snapshot_deletion_failures                    | gauge     | 1           | Snapshot deletion failures by policy
+| elasticsearch_slm_stats_total_scrapes                                 | counter   | 0           | Number of scrapes for SLM collector
+| elasticsearch_slm_stats_json_parse_failures                           | counter   | 0           | JSON parse failures for SLM collector
+| elasticsearch_slm_stats_retention_runs_total                          | counter   | 0           | Total retention runs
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -221,6 +223,23 @@ Further Information
...
 | elasticsearch_clusterinfo_up                                          | gauge     | 1           | Up metric for the cluster info collector
 | elasticsearch_clusterinfo_version_info                                | gauge     | 6           | Constant metric with ES version information as labels
+| elasticsearch_slm_stats_up                                            | gauge     | 0           | Up metric for SLM collector
+| elasticsearch_slm_stats_total_scrapes                                 | gauge     | 0           | Number of scrapes for SLM collector
+| elasticsearch_slm_stats_json_parse_failures                           | gauge     | 0           | JSON parse failures for SLM collector
+| elasticsearch_slm_stats_retention_runs                                | gauge     | 0           | Total retention runs
+| elasticsearch_slm_stats_retention_failed                              | gauge     | 0           | Total failed retention runs
+| elasticsearch_slm_stats_retention_timed_out                           | gauge     | 0           | Total retention run timeouts
```
sysadmind, line 231: These metrics look like counts to me. They should carry the `_total` suffix as per the prometheus standards https://prometheus.io/docs/practices/naming/

README.md
```diff
@@ -221,6 +223,23 @@ Further Information
...
+| elasticsearch_slm_stats_retention_retention_deletion_time_millis      | gauge     | 0           | Retention run deletion time
```
sysadmind, line 232: This has retention twice, is that intentional? Also time metrics should be converted to seconds.

README.md
```diff
@@ -221,6 +223,23 @@ Further Information
...
+| elasticsearch_slm_stats_total_snapshots_taken                         | gauge     | 0           | Total snapshots taken
```
sysadmind, line 233: Many of these also look like counts, not guages.

sysadmind (changes requested):

README.md
Evesy: Thanks @sysadmind -- You're right these are definitely counters so I'll update the docs accordingly. 

It had looked to me that some of the existing counters weren't suffixed with `_total` and had opted instead to keep parity with the name returned in the ES API response, e.g. `elasticsearch_indices_request_cache_evictions` (counter) coming from `request_cache.evictions`, but I'm happy to update the ones I've added to follow best practices for counters

Evesy:
Thanks for your review @sysadmind 

With regards to counter suffix what's your opinion around the below?

Currently there's e.g. `total_snapshots_taken` (Copied verbatim from the ES API response) which represents that value across all policies, and `snapshots_taken` (Also copied verbatim) which is labelled by policy.

Adding a `_total` suffix to these to align with best practices for naming counters would yield:
`total_snapshots_taken_total` & `snapshots_taken_total` which is nice as it still aligns with the ES API but feels like a mouthful.
Another option would be to go away from keeping parity with the API response and instead try and name these more descriptively to more clearly call out one is for overall totals and the other for policies?

There's also an argument that some of the total metrics could be removed in favour of just the policy metrics (as you can just `sum` them to get the same results they currently give), but for some users in large clusters they may not want the cardinality of per policy metrics and choose instead to just keep the totals in their scrape config.

Not sure what's best here? I'm leaning towards the first option of e.g. `total_snapshots_taken_total` & `snapshots_taken_total`

Evesy:
Made the relevant metrics counters now and suffixed appropriately, happy to review/change if you have any differing thoughts on the above though

collector/slm_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+func TestSLM(t *testing.T) {
+	// Testcases created using:
+
+	//  docker run -d -p 9200:9200 -e discovery.type=single-node -e path.repo=/tmp/backups docker.elastic.co/elasticsearch/elasticsearch:7.15.0-arm64
+	//  curl -XPUT http://127.0.0.1:9200/_snapshot/my_repository -H 'Content-Type: application/json' -d '{"type":"url","settings":{"url":"file:/tmp/backups"}}'
+	//  curl -XPUT http://127.0.0.1:9200/_slm/policy/everything -H 'Content-Type: application/json' -d '{"schedule":"0 */15 * * * ?","name":"<everything-{now/d}>","repository":"my_repository","config":{"indices":".*","include_global_state":true,"ignore_unavailable":true},"retention":{"expire_after":"7d"}}'
+	//  curl http://127.0.0.1:9200/_slm/stats (Numbers manually tweaked)
+
+	tcs := map[string][]string{
```
sysadmind, line 34: I think this would be more readable if it was a `map[string]testData` with a new `type testData struct {}`. That would mean instead of the array being positional in the test server, the actual response could be identified by name.

collector/slm.go
```diff
@@ -0,0 +1,401 @@
+// Copyright 2021 The Prometheus Authors
```
sysadmind, line 1:
```suggestion
// Copyright 2022 The Prometheus Authors
```
This needs to be updated in all of the new files.

sysadmind (changes requested):
Thanks for the updates. Just a few minor changes. Otherwise, looks good!

collector/slm_test.go
Evesy: I've just made this into a string rather than an array as the tests only use a single fixture as it stands, and I can't see there needing to be additional fixtures for the same Elastic version with this current collector, however it could always be introduced in the future if needed

Does that feel OK to you?

sysadmind (approved):

PR #575, opened 2022-05-30
feat: add support to elasticsearch v2.x

dougppaz:
- add elasticsearch 2.4.6 nodestats fixture
- load node stats `os.load_average` as `os_load1` gauge
- load node stats `os.cpu_percent` as `os_cpu_percent` gauge
- add indice suggest (total, time and current) from node stats


SuperQ:
In general, we don't expose OS stats from systems like this. Since they overlap with metrics from the node_exporter or cAdvisor, depending on the runtime. While things like ElasticSearch expose these metrics, we don't forward them on to Prometheus.

PR #585, opened 2022-06-13
Support Env var overrides

amexboy:
To ease configuration of the exporter running in K8s and other container environments, environmental variables provide an easier means to change configurations. 

kingpin already supports Envars.


bjk-reddit:
We don't use env vars this way in Prometheus exporters. Kubernetes fully supports args, which are just as easy and more flexible than env vars. For example, you can't do repeated values in ENV vars.

PR #596, opened 2022-07-01
Support for AWS Elasticsearch using AWS SDK v2

### LINKED ISSUE #591: Support for AWS Elasticsearch using AWS SDK v2
Problem: Lacking AWS IAM support to authenticate with Elasticsearch.
Solution: Use AWS SDK v2 to create an IAM session and implement an HTTP round-tripper that will sign requests using Amazon's Signature Verification V4 signing procedure, using credentials from the default AWS credential chain.
To implement this - an extra optional flag is needed to determine the Elasticsearch region i.e. `aws.region`
if the region is present then only the AWS IAM support will work.

rosspatil:
fixes #591


rosspatil:
Hey @sysadmind & @SuperQ,

I have added support for AWS-based elasticsearch authentication using AWS SDK v2. I saw other PR opened for the same but I wanted to add support using SDK v2 because it provides enormous benefits related to how we use the authentication. It supports out-of-the-box all tye of AWS auth mechanisms like env Variable, shared file or web toke file in K8s.

Could you please review the PR and let me know if anything is not appropriate? 

Thanks,
Roshan :)

IBota (approved):

PR #597, opened 2022-07-01
Support for AWS Elasticsearch using AWS SDK v2

### LINKED ISSUE #591: Support for AWS Elasticsearch using AWS SDK v2

rosspatil:
fixes #591 
Hey @sysadmind & @SuperQ,

I have added support for AWS-based elasticsearch authentication using AWS SDK v2. I saw other PR opened for the same but I wanted to add support using SDK v2 because it provides enormous benefits related to how we use the authentication. It supports out-of-the-box all type of AWS auth mechanisms like env Variable, shared file or web toke file in K8s.

Could you please review the PR and let me know if anything is not appropriate?

Thanks,
Roshan :)


SuperQ:
Hmm, for some reason, circleci is not running.

rosspatil:
> Hmm, for some reason, circleci is not running.



> Hmm, for some reason, circleci is not running.

@SuperQ I am thinking maybe because the base is my fork, not the original one. ?

rosspatil:
@SuperQ I found that forked PRs are not considered trusted by default in Circle CI. To make it happen admin needs to perform certain action mentioned in the images. Source of information - https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks. 

Could you please allow running Circle CI on forked PRs?
image

pkg/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+type AWSSigningTransport struct {
+	t      http.RoundTripper
+	creds  aws.Credentials
+	region *string
+	log    log.Logger
+}
+
+func NewAWSSigningTransport(transport http.RoundTripper, region *string, log log.Logger) *AWSSigningTransport {
```
sysadmind, line 45: Why is region a pointer? It's not used as a pointer later.

pkg/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,96 @@
...
+	cfg, err := config.LoadDefaultConfig(context.Background(), config.WithRegion(*region))
+	if err != nil {
+		_ = level.Error(log).Log("msg", "fail to load aws default config", "err", err)
```
sysadmind, line 48: This should return an error, not `os.Exit`. The calling location can choose to exit.

pkg/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,96 @@
...
+		os.Exit(1)
+	}
+
+	creds, err := cfg.Credentials.Retrieve(context.Background())
+	if err != nil {
+		_ = level.Error(log).Log("msg", "fail to retrive aws credentials", "err", err)
+		os.Exit(1)
```
sysadmind, line 55: Same here. This function shouldn't call `os.Exit`

pkg/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return a.t.RoundTrip(req)
+}
+
+func hashPayload(r io.ReadCloser) (payloadHash string, newReader io.ReadCloser, err error) {
+	var payload []byte
+	if r == nil {
+		payload = []byte("")
+	} else {
+		payload, err = ioutil.ReadAll(r)
```
sysadmind, line 87: `r` needs to be closed.

sysadmind (changes requested):

SuperQ:
@rosspatil build forks has been enabled for this repo since the beginning.

carole-lavillonniere (approved):

pkg/roundtripper/roundtripper.go
rosspatil: Removed the pointer

pkg/roundtripper/roundtripper.go
rosspatil: fixed

pkg/roundtripper/roundtripper.go
rosspatil: Fixed

pkg/roundtripper/roundtripper.go
rosspatil: now it will be closed.

rosspatil:
> @rosspatil build forks has been enabled for this repo since the beginning.

@SuperQ Now it seems to be running :)

rosspatil:
@SuperQ @sysadmind Could you please review this PR again? :)

pkg/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,96 @@
...
+	req.Body = newReader
+	err = signer.SignHTTP(context.Background(), a.creds, req, payloadHash, service, a.region, time.Now())
+	if err != nil {
+		_ = level.Error(a.log).Log("msg", "fail to sign request body", "err", err)
+		return nil, err
+	}
+	return a.t.RoundTrip(req)
+}
+
+func hashPayload(r io.ReadCloser) (payloadHash string, newReader io.ReadCloser, err error) {
```
SuperQ, line 81: Please use unnamed returns to avoid using naked returns.

SuperQ (changes requested):

pkg/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,99 @@
...
+	if r == nil {
+		payload = []byte("")
+	} else {
+		defer r.Close()
+		payload, err := ioutil.ReadAll(r)
+		if err != nil {
+			return "", newReader, err
+		}
+		newReader = ioutil.NopCloser(bytes.NewReader(payload))
+	}
```
SuperQ, line 95: This could be simplified a bit.
```suggestion
	var newReader io.ReadCloser
	payload := []byte("")
	if r != nil {
		defer r.Close()
		payload, err := ioutil.ReadAll(r)
		if err != nil {
			return "", newReader, err
		}
		newReader = ioutil.NopCloser(bytes.NewReader(payload))
	}
```

pkg/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,99 @@
+// Copyright 2021 The Prometheus Authors
```
SuperQ, line 1:
```suggestion
// Copyright 2022 The Prometheus Authors
```

SuperQ (approved):

sysadmind (approved):

PR #599, opened 2022-07-04
feat: add indices health collector

dougppaz:


mouseshoes:
Hey Pal! What permissions are required to use these extra collectors?

PR #603, opened 2022-07-15
Fix cluster settings collection when max_shards_per_node is manually set

### LINKED ISSUE #509: Cluster settings collection fails when cluster.max_shards_per_node is set
Hi,

we are experiencing an issue on our 7.14 clusters that after setting cluster.max_shards_per_node manually the collector failed to unmarshal it.
```
cluster_settings.go:160 msg="failed to fetch and decode cluster settings stats" err="json: cannot unmarshal object into Go struct field Cluster.defaults.cluster.max_shards_per_node of type string"
```

The version we are using is 1.2.1, yet the code responsible for this part didn't seem to be changed between 1.2.1 to 1.3.0.

After some investigation we found that

- Without cluster.max_shards_per_node being set manually, the response json from /_cluster/settings?include_defaults would be like
  ``` json
  {
    "defaults": {
      "cluster": {
        "max_shards_per_node.frozen": "3000",
        "max_shards_per_node": "1000"
    }
  }
  ```
- With cluster.max_shards_per_node being set manually as a persistent setting, the response json would become
  ``` json
  {
    "persistent": {
      "cluster": {
        "max_shards_per_node": "1000"
      }
    },
    "defaults": {
      "cluster": {
        "max_shards_per_node": {
          "frozen": "3000"
        }
      }
    }
  }
  ```

This seems to be the reason for this error.
Any ideas?

akazs:
Signed-off-by: akazs <5749172+akazs@users.noreply.github.com>

This PR fixes #509.

### DIFF
--- a/collector/cluster_settings.go
+++ b/collector/cluster_settings.go
@@ -174,8 +174,10 @@ func (cs *ClusterSettings) Collect(ch chan<- prometheus.Metric) {
 
 	cs.shardAllocationEnabled.Set(float64(shardAllocationMap[csr.Cluster.Routing.Allocation.Enabled]))
 
-	maxShardsPerNode, err := strconv.ParseInt(csr.Cluster.MaxShardsPerNode, 10, 64)
-	if err == nil {
-		cs.maxShardsPerNode.Set(float64(maxShardsPerNode))
+	if maxShardsPerNodeString, ok := csr.Cluster.MaxShardsPerNode.(string); ok {
+		maxShardsPerNode, err := strconv.ParseInt(maxShardsPerNodeString, 10, 64)
+		if err == nil {
+			cs.maxShardsPerNode.Set(float64(maxShardsPerNode))
+		}
 	}
 }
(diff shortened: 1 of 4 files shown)


sysadmind (commented):
Is there a way to generate a test for the alternative case where max_shards_per_node is a JSON object?

akazs:
@sysadmind 
It's inside `settings-persistent-clustermaxshartspernode-7.17.5.json` under the default settings. This always exists (at least in 7.x and 8.x to date) when `cluster.max_shards_per_node` and `cluster.max_shards_per_node.frozen` are not in the same part of the settings (default, transient, persistent).

```
$ cat settings-persistent-clustermaxshartspernode-7.17.5.json | jq '.defaults.cluster."max_shards_per_node"'
{
  "frozen": "3000"
}
```

Should I also generate a case putting it in the persistent settings like this?

```json
  "persistent": {
    "cluster": {
      "max_shards_per_node": {
        "frozen": "3000"
      }
    }
  },
```

sysadmind (approved):
LGTM. Thanks for the clarification. I missed the object further down.

Issue #310, opened 2019-12-17
Feature request - Watermark and read_only indices


frenkye:
Hi,

great exporter, but Iam missing variable for watermark and percent usage of node.

1) Node: ```curl localhost:9200/_cat/allocation?h=node,disk.percent```
2) Watermark: ```curl localhost:9200/_cluster/settings``` if not present in persistent or transient settings, then 85% is defaut.

More about watermark is here https://www.elastic.co/guide/en/elasticsearch/reference/7.5/disk-allocator.html

Because if ES hit the flood_stage (introduce in version 6, before high was the limit), then all indices are switch to read only and have to be manually unlock AFAIK, they planned to do auto unlocking in disk space is freed, but don't know if that is implemented in new versions.

With this would be also helpfull if there was variable for tracking number of read only indices, this can be found under ```/_all/_settings``` if set ```index.blocks.read_only_allow_delete: true```

For example like was issue here https://github.com/elastic/kibana/issues/13685

stefnestor:
> Because if ES hit the flood_stage (introduce in version 6, before high was the limit), then all indices are switch to read only and have to be manually unlock AFAIK, they planned to do auto unlocking in disk space is freed, but don't know if that is implemented in new versions.

Confirming from Elastic's [flood-stage troubleshooting guide](https://www.elastic.co/guide/en/elasticsearch/reference/current/fix-watermark-errors.html), the `read_only_allow_delete` block automatically removes after node leaves flood-stage+high watermark thresholds. Implemented v7.4.0 via [elasticsearch#42559](https://github.com/elastic/elasticsearch/pull/42559).

sysadmind:
As for the original request, the `--collector.clustersettings` argument exports cluster settings metrics including watermark settings. i.e.
```
# HELP elasticsearch_clustersettings_allocation_watermark_flood_stage_ratio Flood stage watermark as a ratio.
# TYPE elasticsearch_clustersettings_allocation_watermark_flood_stage_ratio gauge
elasticsearch_clustersettings_allocation_watermark_flood_stage_ratio 0.95
# HELP elasticsearch_clustersettings_allocation_watermark_high_ratio High watermark for disk usage as a ratio.
# TYPE elasticsearch_clustersettings_allocation_watermark_high_ratio gauge
elasticsearch_clustersettings_allocation_watermark_high_ratio 0.9
[...]

PR #611, opened 2022-08-08
add watermark metrics

tomachine:
This solves #310 

Adding 4 new metrics: Disk allocation watermark high, low, flood_stage and disk threshold_enabled.
Metrics support both types of disk allocation watermarks - size in bytes and in percent (shown in label of the metrics).

### DIFF
--- a/collector/cluster_settings.go
+++ b/collector/cluster_settings.go
@@ -20,6 +20,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 
 	"github.com/go-kit/log"
 	"github.com/imdario/mergo"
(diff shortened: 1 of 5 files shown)


lcaflc:
I add support on this feature. This is an highly needed metric for the monitoring of a production cluster. This allow a clean monitoring state of the ES cluster disk usage status without the need to create host dependent disk usage alarms which is painfull with multiple cluster using different watermark levels.

collector/cluster_settings.go
```diff
@@ -21,22 +21,52 @@ import (
...
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+type watermarkMetric struct {
+	Type          prometheus.ValueType
+	Desc          *prometheus.Desc
+	ValueWithUnit func(clusterSettings ClusterSettingsResponse) (float64, string)
+}
+
+func getUnitAndValue(value string) (float64, string) {
```
sysadmind, line 39: I don't see any examples where the values have bytes, only percentage. Do you know which version this changed? 

It is against the best practices to have units be a label. Units should always be normalized (i.e. bytes or seconds). I think it would be best to keep the percentage here as a gauge, but convert percentage to a float from 0-1. 

I'm also not sure about bringing in another dependency to do the unit conversion.

rdegez:
Would love this feature too (at least percentage) so we can have dynamic alerting threshold based on configured wattermarks and not arbitrary static ones.

collector/cluster_settings.go
frenkye: @sysadmind Hi, sorry for delay here. @tomachine prepared this for my issue. We aproach it this way, because this setting could be set both ways and this was sounds acceptable when he wrote this.

If you have other suggestion how to approach this, we are open to suggestions.

You can read about this in official docs and I think this was possible set to percent or absolute value in version 6. Link: https://www.elastic.co/guide/en/elasticsearch/reference/8.6/modules-cluster.html#disk-based-shard-allocation

> cluster.routing.allocation.disk.watermark.low logo cloud
    (Dynamic) Controls the low watermark for disk usage. It defaults to 85%, meaning that Elasticsearch will not allocate shards to nodes that have more than 85% disk used. It can alternatively be set to a ratio value, e.g., 0.85. It can also be set to an absolute byte value (like 500mb) to prevent Elasticsearch from allocating shards if less than the specified amount of space is available. This setting has no effect on the primary shards of newly-created indices but will prevent their replicas from being allocated.

sysadmind: Reading the docs, I can see where this value may be set to a byte value. The docs also mention that the value could be a ratio like `0.8` so that should also be considered. 

Looking over this code again, I think overall it is probably pretty close to where it needs to be, however I'm not a fan of the added 3rd party dependency. The package does not seem to be widely used and it means this exporter has to depend on an outside party for another package.  I'm also not sure, based on the tests in that package that it would properly parse a value like `500mb` anyway.

tomachine:
I rebased and removed the 3rd party dependency. Also solved the issue with value in bytes/ratio/percents watermark by two types of metrics - that will be collected according to type of watermark value.

collector/cluster_settings.go
```diff
@@ -151,5 +208,106 @@ func (c *ClusterSettingsCollector) Update(ctx context.Context, ch chan<- prometh
...
+	ch <- prometheus.MustNewConstMetric(
+		clusterSettingsDesc["lowRatio"],
+		prometheus.GaugeValue,
+		getValueAsRatio(merged.Cluster.Routing.Allocation.Disk.Watermark.Low),
+	)
+
 	return nil
 }
+
+func getValueInBytes(value string) float64 {
```
sysadmind, line 267: As a followup, this should return an error so that the metric is not reported and an error is logged when we fail to parse. There are also no tests for this function or the one below to make sure that no future changes break this functionality.

sysadmind (approved):

PR #614, opened 2022-08-19
Update to Go 1.19

Sijoma:
This bumps the Elasticsearch exporter to Go 1.19. I tried to find all places that reference either Go 1.17 or Go 1.18.

It seems like it would work just fine with the latest version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/prometheus-community/elasticsearch_exporter
 
-go 1.17
+go 1.19
 
 require (
 	github.com/aws/aws-sdk-go-v2 v1.16.7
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/prometheus-community/elasticsearch_exporter
 
-go 1.17
+go 1.19
```
sysadmind, line 3: This one doesn't need to change for a go version bump although I think bumping it to at least 1.18 improves build dependency information. In this case I think it's fine unless @SuperQ has any objections.

sysadmind (approved):
LGTM. Thanks

PR #615, opened 2022-08-23
Add https to metrics server

yuriykis:


tls-config.yml
```diff
@@ -0,0 +1,3 @@
+tls_server_config:
```
sysadmind, line 1: I don't think that this file is necessary.

main.go
```diff
@@ -53,7 +54,8 @@ func main() {
 		metricsPath = kingpin.Flag("web.telemetry-path",
 			"Path under which to expose metrics.").
 			Default("/metrics").String()
-		esURI = kingpin.Flag("es.uri",
+		tlsConfigFile = kingpin.Flag("web.config", "Path to config yaml file that can enable TLS").Default("").String()
```
sysadmind, line 57: This should be handled by the kingpinflag module from the exporter-toolkit.
`webflag "github.com/prometheus/exporter-toolkit/web/kingpinflag"`

```suggestion
		webConfig = webflag.AddFlags(kingpin.CommandLine)
```

main.go
```diff
@@ -288,11 +290,8 @@ func main() {
 	)
 
 	go func() {
-		if err := server.ListenAndServe(); err != nil {
-			_ = level.Error(logger).Log(
-				"msg", "http server quit",
-				"err", err,
-			)
+		if err := web.ListenAndServe(server, *tlsConfigFile, logger); err != nil {
```
sysadmind, line 293:
```suggestion
		if err := web.ListenAndServe(server, *webConfig, logger); err != nil {
```

sysadmind (changes requested):
Just a small adjustment to the kingpin flags. Otherwise good. Thanks!

yuriykis (commented):
Yes, the requested changes look good to me :)

sysadmind (approved):
LGTM. Thanks

Issue #572, opened 2022-05-23
how to get ingest metrics


feiyuyitiao:
use  GET /_nodes/stats/ingest,  I can get  
---
      "ingest" : {
        "total" : {
          "count" : 8670962032,
          "time_in_millis" : 1152040490,
          "current" : 1,
          "failed" : 274
        },
---- 
but from elasticsearch_exporter , I can't find ingest related metrics. 

I check the code ,
 file: nodes.go,  
method  func (c *Nodes) Collect(ch chan<- prometheus.Metric) {  ..........

It has metrics of ( GC Stats | Breaker stats| Thread Pool stats .... ) , but didn't have  ingest metrics implement. 
I think ingest metrics is very useful , can we support?  thanks!

Pyouba:
This is good to have, monitoring ingest pipelines for optimizations and discovering bottlenecks.

PR #617, opened 2022-08-28
add ingest total metrics

zengxilong:
It's PR for #572


zengxilong:
@sysadmind Could you please review this PR ? Thanks.

collector/nodes.go
```diff
@@ -1510,6 +1510,54 @@ func NewNodes(logger log.Logger, client *http.Client, url *url.URL, all bool, no
 				},
 				Labels: defaultNodeLabelValues,
 			},
+			{
+				Type: prometheus.CounterValue,
+				Desc: prometheus.NewDesc(
+					prometheus.BuildFQName(namespace, "ingest", "total_count"),
```
sysadmind, line 1516: Counts should end in `_total`

```suggestion
					prometheus.BuildFQName(namespace, "ingest", "total"),
```

collector/nodes.go
```diff
@@ -1510,6 +1510,54 @@ func NewNodes(logger log.Logger, client *http.Client, url *url.URL, all bool, no
...
+				),
+				Value: func(node NodeStatsNodeResponse) float64 {
+					return float64(node.Ingest.Total.TimeInMillis)
+				},
+				Labels: defaultNodeLabelValues,
+			},
+			{
+				Type: prometheus.CounterValue,
+				Desc: prometheus.NewDesc(
+					prometheus.BuildFQName(namespace, "ingest", "total_current"),
```
sysadmind, line 1540: Is this a counter that always increases? If so `_total` should be at the end
```suggestion
					prometheus.BuildFQName(namespace, "ingest", "current"),
```

zengxilong: In fact, it should be a guage because the data will be zeroed out due to JVM restart. i will fix them.

zengxilong: If the type is Gauge, The "total" prefix should be appropriate？It means the sum of all pipeline stats on a node

zengxilong:
PTAL @sysadmind

rgarcia89:
Where are we with this PR? I would be interested to get to know how much ingest is being passed to my elastic nodes. This way I could build an alert if the ingest increases out of nowhere.

zengxilong:
> Where are we with this PR? I would be interested to get to know how much ingest is being passed to my elastic nodes. This way I could build an alert if the ingest increases out of nowhere.

cc @sysadmind

.gitignore
```diff
@@ -4,4 +4,4 @@ elasticsearch_exporter
 *-stamp
 .tarballs
 /vendor
-.idea
+vendor/
```
SuperQ, line 7: This seems unrelated.

zengxilong: to keep consistent with `.gitignore` of the main branch

Pyouba:
up up 😄

PR #621, opened 2022-09-01
fix health gauge display

dickescheid:
Health gauge was showing `N/A` instead of a meaningful value. Removing the Value from the TableColumn field fixes the issue.
Grafana v9.0.2
Elasticsearch v8.4.0
prometheus_exporter v1.5.0


sysadmind:
This is missing the DCO. Please follow the steps in the DCO check details to fix this.

Issue #624, opened 2022-09-12
AWS Opensearch authentication failure


symmetriac:
We have AWS Opensearch cluster 1.3 and elasticsearch exporter running in EKS cluster with irsa role.
In Opensearch we have role with needed permissions and mapping for the irsa role.
Elasticsearch exporter running with `--aws-region` option so it uses the AWS_WEB_IDENTITY_TOKEN_FILE for authentication.

It's all works well and we see a metrics in grafana and no errors in exporter logs, but after 1 hour it starts to show 403 error code in logs.
In Opensearch audit logs for the exporter requests we can see that it doesn't use irsa role arn anymore:
```
"audit_request_effective_user": "<NONE>"
```

After exporter pod restarting errors are gone and metrics available again, but the issue reoccurs after 1 hour.

symmetriac:
UPD: If I run exporter on ec2 with IAM role attached, it works as expected.

sysadmind:
@symmetriac Is your problem resolved? I'm not sure I understand the original problem, but your last comment makes it sound like the problem has been resolved. If not, please re-open with some more context if possible.

PR #634, opened 2022-10-07
fix: refresh AWS credentials if marked as expired

hellopatrick:
https://github.com/prometheus-community/elasticsearch_exporter/pull/597 adds support for AWS OpenSearch instances, but the credentials may expire if using the EC2 metadata endpoint (IMDS) to fetch credentials.  

As such, before sending request, we should check if the credentials have expired and refresh if need be.


hellopatrick:
I believe this also fixes https://github.com/prometheus-community/elasticsearch_exporter/issues/624, which I'm experiencing as well.  After a certain amount of time, the temporary credentials fetched from IMDS expire and just get 403s afterwards.

hellopatrick:
Hi, @sysadmind, would it be possible get a look/review? Thanks!

PR #314, opened 2020-01-09
Add HTTP/TCP connections and Swap/Recovery metrics

AdrienChevrot1A:
Added the following metrics:

- `elasticsearch_indices_recovery_current_as_source`
- `elasticsearch_indices_recovery_current_as_target`
- `elasticsearch_indices_recovery_throttle_time_seconds_total`
- `elasticsearch_os_swap_in_bytes_used`
- `elasticsearch_os_swap_in_bytes_free`
- `elasticsearch_os_swap_in_bytes_total`
- `elasticsearch_transport_tcp_connections_opened_total`
- `elasticsearch_http_connections_opened_current`
- `elasticsearch_http_connections_opened_total`

PR #635, opened 2022-10-14
Add HTTP metrics and TCP

toufreach5:
Hello,
This PR is inspired from https://github.com/prometheus-community/elasticsearch_exporter/pull/314(author never respond to question)

It will add following metrics:

- elasticsearch_transport_tcp_connections_open_current
- elasticsearch_transport_outbound_connections_total
- elasticsearch_http_connections_opened_current
- elasticsearch_http_connections_opened_total

Thanks.


collector/nodes_response.go
```diff
@@ -37,7 +37,7 @@ type NodeStatsNodeResponse struct {
 	ThreadPool       map[string]NodeStatsThreadPoolPoolResponse `json:"thread_pool"`
 	JVM              NodeStatsJVMResponse                       `json:"jvm"`
 	Breakers         map[string]NodeStatsBreakersResponse       `json:"breakers"`
-	HTTP             map[string]interface{}                     `json:"http"`
+	HTTP             *NodeStatsHTTPResponse                     `json:"http"`
```
sysadmind, line 40: Why is this a pointer?

toufreach5: In order to change this condition:
if len(node.HTTP) == 0 {
to if node.HTTP == nil {

toufreach5: Hello, @sysadmind, any update/concern on this ?

PR #639, opened 2022-10-25
Add cluster label to shard metric

arpitjindal97:
Almost all the other metrics have this label, so it becomes a norm to have it on this metric as well

### DIFF
--- a/collector/shards.go
+++ b/collector/shards.go
@@ -10,11 +10,13 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
 package collector
 
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/prometheus-community/elasticsearch_exporter/pkg/clusterinfo"
 	"net/http"
 	"net/url"
 	"path"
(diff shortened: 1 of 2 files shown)


arpitjindal97:
@sysadmind Can you please have a look at this?

collector/shards.go
```diff
@@ -156,20 +159,26 @@ func (s *Shards) Collect(ch chan<- prometheus.Metric) {
 	nodeShards := make(map[string]float64)
 
 	for _, shard := range sr {
-		if val, ok := nodeShards[shard.Node]; ok {
-			nodeShards[shard.Node] = val + 1
-		} else {
-			nodeShards[shard.Node] = 1
+		if shard.State == "STARTED" {
```
sysadmind, line 162: This seems to change the behavior of the metric. What is the intention here?

sysadmind (commented):
There are a lot of metrics that don't include the cluster name. I'm not sure that it makes sense to create a `Nodes{}` object and pull that data again just to add a label. I think to accept this we would need a better way to share that data.

arpitjindal97:
So, we have like 100 of ElasticSearch clusters and we are running this exporter in each node and collecting metrics in a single prometheus storage.

We want `cluster` label so that we can distinguish between multiple clusters. We have one grafana dashboard with `cluster` as a variable which we are passing in queries. This specific metric doesn't have it, so I raised the Pull Request.

collector/shards.go
arpitjindal97: Ideally, this metrics should be able to tell how many shards each node is handling. The api which this exporter hit is `_cat/shards` which also lists down the relocating shards.

image

Those entries are disruptive and according to existing source code, `node` label gets that whole big string with an arrow and ip address which doesn't make sense.

So, It's better to only capture already relocated shards aka (STARTED)

sysadmind:
I'm on board with this change, but I would like to remove the dependency on `Nodes` to get the cluster name. There is some existing work to have collectors receive updates to the cluster info: https://github.com/prometheus-community/elasticsearch_exporter/blob/master/main.go#L208. It's not the prettiest functionality, but it will cut down on API calls for those user that care about that type of thing. It would also mean if we refactor in the future, we have a common set of functionality to consider.

arpitjindal97:
@sysadmind I have updated the code. Please review

arpitjindal97:
@SuperQ  Can you please review it?

collector/shards.go
```diff
@@ -38,52 +40,100 @@ var (
...
-	Labels func(node string) []string
+	Labels labels
 }
 
 // NewShards defines Shards Prometheus metrics
 func NewShards(logger log.Logger, client *http.Client, url *url.URL) *Shards {
-	return &Shards{
+
+	nodeLabels := labels{
+		keys: func(...string) []string {
```
sysadmind, line 81: Why is this a func if returns a static value? Why does it take parameters if it ignores them?

collector/shards.go
```diff
@@ -38,52 +40,100 @@ var (
...
+			return []string{"node", "cluster"}
```
SuperQ, line 82: Can we name this label `elastic_cluster`? `cluster` is commonly used as an external label.

SuperQ (changes requested):

main.go
```diff
@@ -202,13 +202,18 @@ func main() {
 	prometheus.MustRegister(collector.NewNodes(logger, httpClient, esURL, *esAllNodes, *esNode))
 
 	if *esExportIndices || *esExportShards {
-		prometheus.MustRegister(collector.NewShards(logger, httpClient, esURL))
+		sC := collector.NewShards(logger, httpClient, esURL)
+		prometheus.MustRegister(sC)
 		iC := collector.NewIndices(logger, httpClient, esURL, *esExportShards, *esExportIndexAliases)
 		prometheus.MustRegister(iC)
-		if registerErr := clusterInfoRetriever.RegisterConsumer(iC); registerErr != nil {
+		if clusterInfoRetriever.RegisterConsumer(iC) != nil {
```
sysadmind, line 209: This isn't idiomatic go. 
```suggestion
		if registerErr := clusterInfoRetriever.RegisterConsumer(iC); registerErr != nil {
```

main.go
```diff
@@ -202,13 +202,18 @@ func main() {
...
 			level.Error(logger).Log("msg", "failed to register indices collector in cluster info")
 			os.Exit(1)
 		}
+		if clusterInfoRetriever.RegisterConsumer(sC) != nil {
```
sysadmind, line 213: again, this isn't idiomatic
```suggestion
		if registerErr := clusterInfoRetriever.RegisterConsumer(sC); registerErr != nil {
```

sysadmind (changes requested):

collector/shards.go
arpitjindal97: It's done like this in indices.go

collector/shards.go
arpitjindal97: All the other metrics have cluster label, If we wish to change it to `elastic_cluster` then it should be done in all metrics

sysadmind:
Your comment/screenshot is for `_cat/shards`, but the exporter doesn't use this URL in that way. Your screenshot shows data in a formay that the exporter doesn't see. The URL that the exporter uses is `_cat/shards?format=json`

collector/shards.go
arpitjindal97: In the JSON, I see it's showing `UNASSIGNED` shards also. We only should be capturing `STARTED` shards

image

arpitjindal97: if shards are RELOCATING. this is how it appears:
image

arpitjindal97: This metric only intents to tell how many shards each node currently have. Therefore, we must exclude `RELOCATING` & `UNASSIGNED`

arpitjindal97:
@sysadmind I have added the screenshots from correct API.

collector/shards.go
SuperQ: Yea, seems like a good idea to me. I didn't know the other metrics had the same problem. We can do that as a separate PR if you like.

arpitjindal97:
@sysadmind @SuperQ Please review

SuperQ (changes requested):

.gitignore
```diff
@@ -4,3 +4,4 @@ elasticsearch_exporter
 *-stamp
 .tarballs
 /vendor
+/.idea
```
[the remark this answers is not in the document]
arpitjindal97: Fixed

arpitjindal97:
@sysadmind  @SuperQ  Please review and merge

arpitjindal97:
@sysadmind Please review. I'm really blocked by this

arpitjindal97:
@sysadmind @SuperQ Gentle reminder

sysadmind (approved):
LGTM. The remaining feedback is all around the refactoring of collectors that we are working on, so that can be addressed as part of that effort.

Issue #640, opened 2022-10-26
elasticsearch_indices_active_queries is not present


garrington:
We've setup an environment to test out this exporter.  One of the metrics we'd like to monitor doesn't seem to be present:

elasticsearch_indices_active_queries

Here is the config we're using:

/exporter-bin/elasticsearch_exporter/elasticsearch_exporter \
--es.uri="${DOMAIN_ENDPOINT}" \
--aws.region="us-east-1" \
--log.level="${LOG_LEVEL}" \
--es.all \
--es.cluster_settings \
--es.indices \
--es.indices_mappings \
--es.indices_settings \
--es.shards \
--es.snapshots \
--es.timeout="45s"  \
--log.output="stdout" &

I also cannot find this metric referenced anywhere in code.  Has this metric been deprecated?

sysadmind:
It would appear the the README is wrong. The metric name is elasticsearch_search_active_queries. I think we need to fix this.

lynBIT:
I can't find this metric, and I can't find elasticsearch_indices_index_current either. Do you know why

lynBIT:
A new metrics.I can't find elasticsearch_indices_index_current, 
the config we're using:
/exporter-bin/elasticsearch_exporter/elasticsearch_exporter
--es.uri="${DOMAIN_ENDPOINT}"
--es.all
--es.indices
--es.indices_settings
--es.shards
--es.snapshots
I also cannot find this metric referenced anywhere in code. 
Please help me,thank you @sysadmind

sysadmind:
I'm not sure why you don't see these. I just ran a test and I can see the elasticsearch_index_stats_index_current metric. 

Run elasticsearch in a docker container: 
```
docker run --rm -it -p 9200:9200 -e "xpack.security.enabled=false"  -e "discovery.type=single-node" -e "cluster.name=elasticsearch" -e "ES_JAVA_OPTS=-Xms750m -Xmx750m" elasticsearch:7.16.3
```
Run the exporter:
```
elasticsearch_exporter --es.uri="127.0.0.1:9200" --es.all --es.indices
```
Check the metrics from the exporter:
```
curl -s 127.0.0.1:9114/metrics | grep elasticsearch_index_stats_index_current 
# HELP elasticsearch_index_stats_index_current The number of documents currently being indexed to an index
# TYPE elasticsearch_index_stats_index_current gauge
[...]

PR #650, opened 2022-11-17
fix name of metric

gempir:
In #640 it was mentioned that the metric does not appear in the output. The correct name of the metric is now in the readme.

Also sorted the metric up so it's alphabetical.


SuperQ (approved):

SuperQ:
This needs a rebase.

Issue #400, opened 2021-01-04
Feature Request: add support for AWS IAM Roles for Service Account


oridool:
Hi,
I'd like the exporter to connect to a managed AWS Elasticsearch, and to assume IAM Role rather than specify a username+password. I've configured the cluster to work with IAM roles rather than users.
If the POD is running in EKS, AWS credentials can be fetched using IAM Roles for Service Account (IRSA):
https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html 

Is it possible to add such support to this project?

Thank you.

fire2:
any workaround for this? I don't want the K8S pod to run with one role so that is just assumes another role. It would be really cool if IRSA support would be built in.

PR #636, opened 2022-10-17
Fix AWS credential expiration

sysadmind:
Use the AWS cached credential provider to automatically handle credentials. The CredentialsCache will automatically handle refreshing expired credentials and keeping them cached as long as necessary.

Replaces #634 as this offloads more of the work to the AWS SDK

Signed-off-by: Joe Adams <github@joeadams.io>

Issue #385, opened 2020-09-09
AWS authentication with IAM roles for elasticsearch exporter


Shravankumark:
Hello,

I am using elasticsearch exporter to pull metrics from VPC based AWS elastic search into prometheus.
We are restricting AWS access policy to IAM roles(instance profiles).Does elastic search exporter work well with IAM roles for authentication.
I tried to troubleshoot by installing aws-cli,but the elasticsearch exporter container doesn't allow to install any external packages.
Could you someone help on this?

{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": [
          "arn:aws:iam::123456789:role/test1_role",
          "arn:aws:iam::123456789:role/test2-role"
        ]
      },
      "Action": [
        "es:ESHttpHead",
        "es:ListDomainNames",
        "es:DescribeElasticsearchDomain",
        "es:ESHttpPost",
        "es:ESHttpGet",
        "es:ESHttpPatch",
        "es:DescribeElasticsearchDomains",
        "es:ESHttpDelete",
        "es:ESHttpPut"
      ],
      "Resource": "arn:aws:es:us-east-1:123456789:domain/test-es/*"
    }
  ]
}

PR #653, opened 2022-11-18
Add support to assume an AWS role and renew expired credentials

steveteuber:
Hi, we have added support to assume an AWS role which is required to get access from a different account.
By the way, I think the following issues can be closed: #353, #385, #400, #443, #636

### DIFF
--- a/main.go
+++ b/main.go
@@ -125,6 +125,9 @@ func main() {
 		awsRegion = kingpin.Flag("aws.region",
 			"Region for AWS elasticsearch").
 			Default("").String()
+		awsRoleArn = kingpin.Flag("aws.role-arn",
+			"Role ARN of an IAM role to assume.").
+			Default("").String()
 	)
 
 	kingpin.Version(version.Print(name))
@@ -174,7 +177,7 @@ func main() {
 	}
 
 	if *awsRegion != "" {
-		httpClient.Transport, err = roundtripper.NewAWSSigningTransport(httpTransport, *awsRegion, logger)
+		httpClient.Transport, err = roundtripper.NewAWSSigningTransport(httpTransport, *awsRegion, *awsRoleArn, logger)
 		if err != nil {
 			_ = level.Error(logger).Log("msg", "failed to create AWS transport", "err", err)
 			os.Exit(1)
--- a/pkg/roundtripper/roundtripper.go
+++ b/pkg/roundtripper/roundtripper.go
@@ -26,6 +26,8 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	v4 "github.com/aws/aws-sdk-go-v2/aws/signer/v4"
 	"github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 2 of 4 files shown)


corford:
Would be great if this landed as we're hitting the limitation right now (OpenSearch in a VPC, IAM role grants in the domain's access policy, EKS cluster)

ralbertha (commented):
We'd appreciate it if we can have this feature sooner ;)

steveteuber:
Hey @sysadmind,

Happy New Year! Could we merge this pull request or do you miss something?

DCUKWS:
really would appreciate this being merged please

DCUKWS (approved):
looking to have this merged ASAP

ralbertha (approved):

main.go
```diff
@@ -120,7 +120,10 @@ func main() {
 			"Sets the log output. Valid outputs are stdout and stderr").
 			Default("stdout").String()
 		awsRegion = kingpin.Flag("aws.region",
-			"Region for AWS elasticsearch").
```
sysadmind, line 123: I don't think that this text needs to change. The region is still for the AWS cluster.

pkg/roundtripper/roundtripper.go
```diff
@@ -36,24 +38,24 @@ const (
...
 
-	creds, err := cfg.Credentials.Retrieve(context.Background())
-	if err != nil {
-		_ = level.Error(log).Log("msg", "fail to retrive aws credentials", "err", err)
-		return nil, err
+	if roleArn != "" {
+		cfg.Credentials = stscreds.NewAssumeRoleProvider(sts.NewFromConfig(cfg), roleArn)
 	}
 
+	creds := aws.NewCredentialsCache(cfg.Credentials)
```
sysadmind, line 57: One thing that this misses from the original is to ensure that the credentials are valid before returning. Can we add that functionality back in? e.x.
```suggestion
	creds := aws.NewCredentialsCache(cfg.Credentials)
	// Run a single fetch credentials operation to ensure that the credentials
	// are valid before returning the transport.
	_, err = cfg.Credentials.Retrieve(context.Background())
	if err != nil {
		_ = level.Error(log).Log("msg", "failed to retrive aws credentials", "err", err)
		return nil, err
	}
```

sysadmind (changes requested):
Thank you for your patience. I have been very busy lately. I think this is very close. Just a few nitpicks. 

Thanks for the PR!

pkg/roundtripper/roundtripper.go
steveteuber: Done.

sysadmind (approved):
LGTM Thanks!

patrickdash:
Hi all, thanks for getting this PR into the exporter! Do we have any plans for cutting a new release for this version?

fire2:
this doesn't resolve https://github.com/prometheus-community/elasticsearch_exporter/issues/400 since it doesn't work natively with IRSA

PR #656, opened 2022-11-27
Refactor cluster settings collector

sysadmind:
Refactor cluster settings collector into new Collector interface. Update tests to use prometheus/testutil to validate the output instead of checking the parsed elasticsearch response.


SuperQ (approved):
Nice

PR #672, opened 2023-01-11
Basic auth for the /metrics endpoint

msirovy:
It is small improvement, but it could help simplify life. Please take a look and give me a feedback or simple improve it.


sysadmind:
Basic auth should be handled by the exporter-toolkit instead of a custom implementation. 

See here: https://github.com/prometheus/exporter-toolkit/tree/master/web

msirovy:
Nice. I've missed it. You don't have mentioned it in the documentation. The flag for enabling base auth is 

```bash
--web.config.file=path.yaml
```

sysadmind:
@msirovy I assume that means this PR can be closed?

msirovy:
Closing

msirovy:
Ive found one problem. latest release in this repository does NOT contain this flag. It seems to me like you don't release new package for a long time.

Issue #674, opened 2023-01-20
elasticsearch_indices_mappings_stats_fields metric doesn't appear to report accurate count


hartfordfive:
# Problem Description 

I've recently enabled the `es.indices_mappings` flag to gain the ability to monitor and alert in the case of index field limits being exceeded.  Unfortunately, I've noticed that the number reported by the exporter does not appear to be accurate.   I've confirmed this by testing with a sample document observing the resulting response from Elasticsearch and then confirming via the [field capabilities API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-field-caps.html).   I've outlined the steps to reproduce this issue below.

# Relevant Software Versions

- Elasticsearch 8.4

# Reproducing the problem

### 1. Create a test index

```
PUT /dummy-data-2023.01.20/
```
response:
```
{
	"acknowledged": true,
	"shards_acknowledged": true,
	"index": "dummy-data-2023.01.20"
}
```

### 2. Confirm index mapping is empty:
```
GET /dummy-data-2023.01.20/_mapping
```
response:
```
{
	"dummy-data-2023.01.20": {
		"mappings": {}
	}
}
```

and also via the field capabilities API:
```
GET /dummy-data-2023.01.20/_field_caps?fields=*&filters=-metadata
```
response:
```
{
	"indices": [
		"dummy-data-2023.01.20"
	],
	"fields": {}
}
```

### 3. Set the max number of fields in the index to 20

Take note that by default, each field has text type and a "keyword" via the "fields" property.

```
PUT /dummy-data-2023.01.20/_settings
{
	"index.mapping.total_fields.limit": 40
}
```
response:
```
{
	"acknowledged": true
}
```

and confirm the updated settings have been applied:
```
GET /dummy-data-2023.01.20/_settings
```
response:
```
{
	"dummy-data-2023.01.20": {
		"settings": {
			"index": {
				"routing": {
					"allocation": {
						"include": {
							"_tier_preference": "data_content"
						}
					}
				},
				"mapping": {
					"total_fields": {
						"limit": "40"
					}
				},
				"number_of_shards": "1",
				"provided_name": "dummy-data-2023.01.20",
				"creation_date": "1674226159773",
				"number_of_replicas": "1",
				"uuid": "pSGTiVyuTOWvb3H38VLrDA",
				"version": {
					"created": "8040399"
				}
			}
		}
	}
}
```

### 4. Index a document that has exactly the max number of fields

```
[...]

PR #675, opened 2023-01-21
Fix index field count

hartfordfive:
This change applies a small modification that adds 1 to the field count prior to recursively counting the other nested fields in the mapping. (Issue #674)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+
+## 1.5.1 / 2022-07-29
```
sysadmind, line 2: Version gets changed when we are ready for a release. 
```suggestion
## master / unreleased
```

VERSION
```diff
@@ -1 +1 @@
-1.5.0
```
sysadmind, line 1: This should not be changed.

sysadmind (changes requested):
You're missing the DCO. The check has instructions for how to fix. 
I'm not sure this takes into account a situation where both type and properties are set. I'm not sure if these are mutually exclusive, but if not these would potentially double count those fields.

hartfordfive:
@sysadmind , thank you for the quick feedback.   I will address those required changes you pointed out.   

As for counting a mapping where both type and property are set, although the type `object` could be technically be present with the `properties` field, it isn't usually defined explicitly (at least based on my experience) as that's the default value as per the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html).
```
You are not required to set the field type to object explicitly, as this is the default value.
```
Having said that, the possibility is a valid edge case that should be accounted for, so I'll go ahead and make the modification.  Thank you for bringing that up.

sysadmind (approved):
LGTM Thanks!

PR #716, opened 2023-05-23
Update README: Add required privileges for es.indices_mappings

ramir-savvy:
I tried setting the permissions as described in the readme, but those, in the suggested change, are missing.
Thanks


ramir-savvy:
@zwopir FYI about this PR - I ran into this issue and I thought it would be nice to update the README file. Thanks

SuperQ (approved):

Issue #638, opened 2022-10-20
elasticsearch_thread_pool_rejected_count no longer returns values (other than 0)


shawnclarke7:
We are using elasticsearch-exporter 1.3.0 with elasticsearch 7.17.6
As of elasticsearch 7.9, **elasticsearch_thread_pool_rejected_count** is no longer a valid metric which we used as a leading indicator of load on the elasticsearch cluster. Elastic Support suggested that a possible replacement would be to use **indexing_pressure.memory.current.combined_coordinating_and_primary_in_bytes** and **indexing_pressure.memory.current.all_in_bytes** as alternatives however these values do not seem to be exposed.

Here is a quote from Elastic Support:
> Starting in Elasticsearch 7.9, we’ve increased Elasticsearch ingestion throughout by removing the fixed ingestion queue of 200 documents and replacing it with a new flexible memory-based queue. Through our internal testing, we’ve been able to drastically improve ingestion throughput in Elasticsearch with this new ingestion [subsystem](https://github.com/elastic/elasticsearch/pull/59487), especially when there are large numbers of clients writing data to Elasticsearch.
> 
> At each stage of the indexing process (coordinating, primary, and replication), the new ingestion subsystem accounts for the number of bytes associated with an operation and [rejects requests](https://github.com/elastic/elasticsearch/pull/58885) if a byte-based limit is exceeded (10% of heap memory by default). This feature is enabled by default in Elasticsearch 7.9, and detailed indexing information is exposed via the [node stats API](https://www.elastic.co/guide/en/elasticsearch/reference/7.9/cluster-nodes-stats.html) (see the ‘indexing_pressure’ sections).

Since this change "" will basically never return a value other than 0 therefore we are looking for a different metric to act as a leading indicator to load on the system outside of using the built-in rules in Kibana. Elastic Support offered this suggestion:

>According to our ES Devs, both are important and you could set your desired threshold when monitoring all_in_bytes.
>
>- indexing_pressure.memory.current.combined_coordinating_and_primary_in_bytes:
reports the number of bytes of heap needed to hold all the documents in the coordinating and primary stages
>
>- indexing_pressure.memory.current.all_in_bytes:
reports bytes from all stages of indexing (coordinating, primary, and replica)
>
Additionally, **indexing_pressure.memory.limit_in_bytes** would be helpful to have too.

If these values are already supported, please let me know in which version of elasticsearch-exporter and what they are called. If they are not currently exposed, could you please add them?

[...]

PR #727, opened 2023-06-09
feat: add indexing pressure

mingregister:
So far, I have not seen the indexing_pressure metrics in the elasticsearch_exporter, so i add it.  https://github.com/prometheus-community/elasticsearch_exporter/issues/638


mingregister:
@SuperQ hello, Could you help to review this pr.  I use this pr in our production, and everything is fine,
* elasticsearch7.10 collects all metrics successfully
* elasticsearch7.7 does not have the indexing_pressure metrics but everything else works fine except can not collect the indexing_pressue metric.

Please let me know if you have any questions or concerns.

collector/nodes.go
```diff
@@ -1781,6 +1789,176 @@ func NewNodes(logger log.Logger, client *http.Client, url *url.URL, all bool, no
 				Labels: defaultFilesystemIODeviceLabelValues,
 			},
 		},
+		indexingPressureMetrics: []*indexingPressureMetric{
+			{
+				Type: prometheus.GaugeValue,
+				Desc: prometheus.NewDesc(
+					prometheus.BuildFQName(namespace, "indexing_pressure", "current_combined_coordinating_and_primary_in_bytes"),
```
SuperQ, line 1796: This seems like a composite metric that is the sume of the `current_coordinating_in_bytes` and `current_primary_in_bytes` metrics. We typically drop these things to save space and cardinality when you can use `foo + bar` in PromQL.

mingregister: thanks for your review.  After I check the document, `current_combined_coordinating_and_primary_in_bytes` or `total_combined_coordinating_and_primary_in_bytes` is not equal to `coordinating + primary`.    So maybe we should keep this?    Here is the document:
```
# https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-stats.html#cluster-nodes-stats-api-response-body-indexing-pressure
combined_coordinating_and_primary_in_bytes: 
(integer) Memory consumed, in bytes, by indexing requests in the coordinating or primary stage. This value is not the sum of coordinating and primary as a node can reuse the coordinating memory if the primary stage is executed locally.
```



Please let me known if you have any questions or concern.

PR #752, opened 2023-07-21
Add extra tests and improve error handling in watermark metrics

sysadmind:
Improve the error handling for watermark metrics. When we have an error, do not report a 0 value for the metric.

Adds tests for parsing the ratio/percentage and the human readable bytes in watermark data.

Functionality originally added in #611


sysadmind:
cc @tomachine

SuperQ (approved):
Thanks

PR #574, opened 2022-05-30
Add Indices filter flag

chapost1:
As can be seen earlier, certain people asked for the option to filter the indices they are exporting.

Especially if you have a lot of them and you need to monitor just certain, the rest of it, is just a waste of resources.

ATM, fetching indices settings/stats/mappings queries all the existing ones using this kind of GET requests:
`GET /_all/_settings`, `GET /_all/_stats` and  `GET /_all/_mappings`

As mentioned in the docs:
https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-get-settings.html

It's possible to just retrieve the ones which match 2 types of queries:

1. Matching using wildcard(*) expression. i.e:
`GET /prefix_*/_stats`

2. Matching a list of indices, using comma:
`GET /first_name,second_name/_stats`

It is possible to mix them both. ie:
`GET /prefix_1_*,prefix_2_*/_stats`

The new added flag:
`es.indices_filter` (default is '_all')

<strong>Note: The flag will be used only if any export indices flag is used.</strong>
To use it, pass an expression to match your use-case indices, default '_all' just like before.
By passing one of the options mentioned above as a value (statement with a wildcard, list with commas), it will be used in the http request to retrieve statistics.

This change would be very appreciated,
Thanks.

PR #764, opened 2023-08-17
Add Indices filter flag

chapost1:
`NOTE:` this is a new PR in behalf of https://github.com/prometheus-community/elasticsearch_exporter/pull/574
as the previous one has DCO issue and after trying to squash commit to sign all of them, it messed the PR so it has been decided to open a new one.

`Original Description:`
As can be seen earlier, certain people asked for the option to filter the indices they are exporting.

Especially if you have a lot of them and you need to monitor just certain, the rest of it, is just a waste of resources.

ATM, fetching indices settings/stats/mappings queries all the existing ones using this kind of GET requests:
`GET /_all/_settings`, `GET /_all/_stats` and  `GET /_all/_mappings`

As mentioned in the docs:
https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-get-settings.html

It's possible to just retrieve the ones which match 2 types of queries:

1. Matching using wildcard(*) expression. i.e:
`GET /prefix_*/_stats`

2. Matching a list of indices, using comma:
`GET /first_name,second_name/_stats`

It is possible to mix them both. ie:
`GET /prefix_1_*,prefix_2_*/_stats`

The new added flag:
`es.indices_filter` (default is '_all')

<strong>Note: The flag will be used only if any export indices flag is used.</strong>
To use it, pass an expression to match your use-case indices, default '_all' just like before.
By passing one of the options mentioned above as a value (statement with a wildcard, list with commas), it will be used in the http request to retrieve statistics.

This change would be very appreciated,
Thanks.

### DIFF
--- a/main.go
+++ b/main.go
@@ -89,6 +89,9 @@ func main() {
 		esExportSLM = kingpin.Flag("es.slm",
 			"Export stats for SLM snapshots.").
 			Default("false").Bool()
+		esIndicesFilter = kingpin.Flag("es.indices_filter",
+			"Indices filter by name for which metrics should be exposed, using prefix with wildcards and/or commas as multi-selection delimiter.").
+			Default("_all").String()
 		esExportDataStream = kingpin.Flag("es.data_stream",
 			"Export stas for Data Streams.").
 			Default("false").Bool()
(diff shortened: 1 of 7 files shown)


sysadmind (changes requested):
I think this is pretty straightforward and overall looks pretty good. Thinking about how this is used, I think it would be better to call this an index selector instead of a filter. It's not filtering out indices, it is selecting which ones to get metrics for. That would make the flat `es.indices_selector` and would mean renaming the variables. Code-wise I think everything is okay.

chapost1:
@sysadmind 

Thanks for the rename suggestion.
I've modified it accordingly, and updated the PR.

chapost1:
@sysadmind
any update?

main.go
```diff
@@ -86,6 +86,9 @@ func main() {
 		esExportSLM = kingpin.Flag("es.slm",
 			"Export stats for SLM snapshots.").
 			Default("false").Bool()
+		esIndicesSelector = kingpin.Flag("es.indices_selector",
+			"Selcted indices by name for which metrics should be exposed, using prefix with wildcards and/or commas as multi-selection delimiter.").
```
sysadmind, line 90:
```suggestion
			"Select indices by name for which metrics should be exposed, using prefix with wildcards and/or commas as multi-selection delimiter.").
```

sysadmind (changes requested):
Just one small fix. Thanks for your patience.

chapost1:
@sysadmind 
Thanks, I've committed your change proposal

Issue #525, opened 2022-01-20
New Task Management API


mrbetta:
Elasticsearch has a new API in beta that shows pending inter-cluster tasks like searches and cleanup activities. These are different than the tasks Elasticsearch Exporter currently exports, which are pending administrative tasks.

https://www.elastic.co/guide/en/elasticsearch/reference/current/tasks.html
https://github.com/elastic/elasticsearch/issues/51628

Would it be possible to add support for this new API so that we can monitor these tasks as another health indicator.

PR #778, opened 2023-09-20
collector: add tasks API collection

devoxel:
This commit adds simple aggregation of Elasticsearch Tasks API.
There are 4 new metrics; though 3 are just bookkeeping.

elasticsearch_task_stats_action_total is a gague reporting the total
number of tasks running for a given action. Because there are no stats
endpoints available for this, this change introduces an aggregation step
to group the number of tasks by action name.

This metric is useful for ensuring long running actions of a specific
kind stay within a specific limit. Of particular use to me is
the action: 'indices:data/write/delete/byquery'.

In my usecase, our ES access patterns mean we have a predefined limit
of these actions running on the cluster.

This change also adds two new CLI flags to manage the collection of tasks API:

	--es.tasks (to enable task collection)
	--es.tasks.actions (to filter tasks by action param)

Issue https://github.com/prometheus-community/elasticsearch_exporter/issues/525 proposed addition of collection of these tasks.


devoxel:
@sysadmind hey - would it be possible to review this?

collector/tasks.go
```diff
@@ -0,0 +1,174 @@
...
+)
+
+// Task Information Struct
+type Task struct {
+	logger  log.Logger
+	client  *http.Client
+	url     *url.URL
+	actions string
+
+	up                              prometheus.Gauge
```
sysadmind, line 47: I think we can skip the bookkeeping metrics. I don't think they add value.

collector/tasks.go
```diff
@@ -0,0 +1,174 @@
...
+		}),
+		totalScrapes: prometheus.NewCounter(prometheus.CounterOpts{
+			Name: prometheus.BuildFQName(namespace, "task_stats", "total_scrapes"),
+			Help: "Current total Elasticsearch snapshots scrapes.",
+		}),
+		jsonParseFailures: prometheus.NewCounter(prometheus.CounterOpts{
+			Name: prometheus.BuildFQName(namespace, "task_stats", "json_parse_failures"),
+			Help: "Number of errors while parsing JSON.",
+		}),
+		byActionMetrics: []*taskByAction{
```
sysadmind, line 73: I don't think we need a slice of metrics here. In fact, I think we can just declare the Desc as a package level var, and we can use it with `prometheus.MustNewConstMetric`. Having named vars for metrics helps with the linter finding unused metrics.

collector/tasks.go
```diff
@@ -0,0 +1,174 @@
...
+
+	ch <- t.up.Desc()
+	ch <- t.totalScrapes.Desc()
+	ch <- t.jsonParseFailures.Desc()
+}
+
+func (t *Task) fetchAndDecodeAndAggregateTaskStats() (*AggregatedTaskStats, error) {
+	u := *t.url
+	u.Path = path.Join(u.Path, "/_tasks")
+	u.RawQuery = "group_by=none&actions=" + t.actions
```
sysadmind, line 106: I'm not sure it's a good idea to use string concatenation here. There is no encoding on t.actions. 

I think we should either use a `URL.Params` and then call `u.RawQuery = params.Encode()` or wrap t.actions in an Encode().

collector/tasks_response.go
```diff
@@ -0,0 +1,39 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+// TasksResponse is a representation of the Task management API.
+type TasksResponse struct {
```
sysadmind, line 17: JSON response definitions should go into the tasks.go file. We don't need a separate file just for this.

collector/tasks_response.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// TaskResponse is a representation of the individual task item returned by task API endpoint.
+//
+// We only parse a very limited amount of this API for use in aggregation.
+type TaskResponse struct {
+	Action string `json:"action"`
+}
+
+type AggregatedTaskStats struct {
```
sysadmind, line 28: I'm not sure that there is a benefit to having this type. We could just pass the map instead.

collector/tasks_test.go
```diff
@@ -0,0 +1,77 @@
...
+		}))
+		defer ts.Close()
+
+		u, err := url.Parse(ts.URL)
+		if err != nil {
+			t.Fatalf("Failed to parse URL: %s", err)
+		}
+
+		task := NewTask(log.NewNopLogger(), http.DefaultClient, u, "indices:*")
+		stats, err := task.fetchAndDecodeAndAggregateTaskStats()
```
sysadmind, line 54: I think it would be better to test against the metric output here.

collector/tasks_test.go
```diff
@@ -0,0 +1,77 @@
...
+		if err != nil {
+			t.Fatalf("Failed to fetch or decode data stream stats: %s", err)
+		}
+		t.Logf("[%s] Task Response: %+v", ver, stats)
+
+		// validate actions aggregations
+		if len(stats.CountByAction) != 4 {
+			t.Fatal("expected to get 4 tasks")
+		}
+		if stats.CountByAction["indices:data/write/index"] != 1 {
```
sysadmind, line 64: I think it would be more meaningful if all of these had different values. As it stands, they could be transposed and the tests would not catch it.

sysadmind (changes requested):
We are moving towards the Collector interface for collectors. If you look at cluster_info and cluster_settings, you will see examples of the newer style. If that was the only change, I would say we could fix it up after merging, but with the other review remarks, I think it's worthwhile to do at the same time. Using the Collector interface means --es.tasks is no longer necessary. It will be --collector.tasks automatically. The --es.tasks.actions flag will still be necessary, but I don't think that the "es." prefix is necessary. 

Overall, I think this is close and the logic seems to be correct.

collector/tasks.go
devoxel: Fixed in migration to collector interface.

collector/tasks.go
devoxel: Fixed (went the former solution)

collector/tasks.go
devoxel: Done

collector/tasks_response.go
devoxel: Done

collector/tasks_test.go
devoxel: Done

collector/tasks_test.go
devoxel: Done

collector/tasks_response.go
devoxel: I think having the struct makes the code a bit more self documenting; it makes it clear we're doing aggregation and by what. If you'd prefer I can make that change though

devoxel:
Thanks for the prompt & useful review @sysadmind :) 

>  The --es.tasks.actions flag will still be necessary

I ended up moving the flag into tasks.go directly and using a global; because the collector factory cannot pass params during construction.

The downside of directly gathering flags in the `init()` are:

- A package level global mutable variable is never nice; some other function could accidentally change the behaviour of tasks

- If the package is imported it's flags are infectious. However, `registerCollector` does this anyway; so I don't think this issue is worth discussion.

I considered some alternatives: 

- Extending Collector interface: this ends up with us either exposing flag implementation details or only supporting string value options

- A struct tag solution with `reflect.TypeOf`: this seems nice on paper but implementation is way too complex for such a small gain

If you have a better implementation please suggest.

collector/tasks.go
```diff
@@ -0,0 +1,144 @@
...
+// We only parse a very limited amount of this API for use in aggregation.
+type TaskResponse struct {
+	Action string `json:"action"`
+}
+
+type AggregatedTaskStats struct {
+	CountByAction map[string]int64
+}
+
+func AggregateTasks(t TasksResponse) *AggregatedTaskStats {
```
sysadmind, line 137: - This doesn't need to be public
- Why the pointer? I don't think it makes sense. 
```suggestion
func aggregateTasks(t TasksResponse) aggregatedTaskStats {
```

collector/tasks.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return tr, err
+	}
+
+	err = json.Unmarshal(bts, &tr)
+	return tr, err
+}
+
+// TasksResponse is a representation of the Task management API.
+type TasksResponse struct {
```
sysadmind, line 122: All of these type structs should be private. I don't see much value in making them public.

collector/tasks.go
devoxel: Fixed.

devoxel:
Hey - @sysadmind - any update here? I have addressed your most recent set of comments

collector/tasks.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/go-kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// filterByTask global required because collector interface doesn't expose any way to take
+// constructor args.
+var actionFilter string
+
+var taskActionDesc = prometheus.NewDesc(
+	prometheus.BuildFQName(namespace, "task_stats", "action_total"),
```
sysadmind, line 35: This should not have a _total suffix because it is not a counter.

sysadmind (changes requested):
Everything looks good except the metric name. Thanks for your patience.

sysadmind (approved):
LGTM

PR #789, opened 2023-10-12
Refactor snapshots collector

sysadmind:
- Move metric Desc to vars to aid in unused linter checks
- Use new Collector interface
- Add a util getURL func. Similar funcs are in most of the collectors. The only difference is typically what struct to unmarshal the JSON into. This func just returns []byte so that the caller can handle the unique data structure.

This does not adjust the labels on the metrics. I want that to be handled separately to make sure the refactor itself matches existing behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## master / unreleased
+
+BREAKING CHANGES:
+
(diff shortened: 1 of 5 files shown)


SuperQ (approved):
Nice

PR #811, opened 2023-11-15
Refactor tests for cluster health

sysadmind:
- Remove up, totalScrapes, and jsonParseFailures metrics. They are not useful.
- Move fixtures to individual files
- Base tests on the metric output for better testing the expected output instead of the internals.


SuperQ (commented):
If `Collect()` doesn't return a 500 error on failure, we probably want to keep the `up` metric.

SuperQ (approved):
Oh, I see, this is a subsystem up. That's not useful.

PR #816, opened 2023-11-15
Add creation_date metrics

capuche2412:
Hello,

This PR has the purpose to add a new metrics :
`elasticsearch_indices_settings_creation_timestamp_seconds`

The motivation was that we want to keep tracks of the index creation dates. We compare it with log timestamp and create an alert if logs are timestamped in the future or in the past

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ Further Information
 | elasticsearch_indices_search_query_total                             | counter    | 1           | Total number of queries                                                                             |
 | elasticsearch_indices_segments_count                                 | gauge      | 1           | Count of index segments on this node                                                                |
 | elasticsearch_indices_segments_memory_bytes                          | gauge      | 1           | Current memory size of segments in bytes                                                            |
-| elasticsearch_indices_settings_creation_timestamp_seconds            | gauge      | 1           | Timestamp of the index creation                                                                     |
+| elasticsearch_indices_settings_creation_timestamp_seconds            | gauge      | 1           | Timestamp of the index creation in seconds                                                                     |
 | elasticsearch_indices_settings_stats_read_only_indices               | gauge      | 1           | Count of indices that have read_only_allow_delete=true                                              |
 | elasticsearch_indices_settings_total_fields                          | gauge      |             | Index setting value for index.mapping.total_fields.limit (total allowable mapped fields in a index) |
 | elasticsearch_indices_settings_replicas                              | gauge      |             | Index setting value for index.replicas                                                              |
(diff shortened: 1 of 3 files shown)


collector/indices_settings.go
```diff
@@ -105,6 +106,21 @@ func NewIndicesSettings(logger log.Logger, client *http.Client, url *url.URL) *I
 					return val
 				},
 			},
+			{
+				Type: prometheus.GaugeValue,
+				Desc: prometheus.NewDesc(
+					prometheus.BuildFQName(namespace, "indices_settings", "creation_date"),
```
SuperQ, line 112: This needs to follow Prometheus naming practices.

```suggestion
					prometheus.BuildFQName(namespace, "indices_settings", "creation_timestamp_seconds"),
```

README.md
```diff
@@ -187,6 +187,7 @@ Further Information
 | elasticsearch_indices_search_query_total                             | counter    | 1           | Total number of queries                                                                             |
 | elasticsearch_indices_segments_count                                 | gauge      | 1           | Count of index segments on this node                                                                |
 | elasticsearch_indices_segments_memory_bytes                          | gauge      | 1           | Current memory size of segments in bytes                                                            |
+| elasticsearch_indices_settings_creation_date                         | gauge      | 1           | Index setting value for index.creation_date                                                         |
```
SuperQ, line 190: This should be more descriptive, and should match the Help text.

Maybe something like "Timestamp of the index creation".

SuperQ (changes requested):

SuperQ (approved):

collector/indices_settings.go
```diff
@@ -105,6 +106,21 @@ func NewIndicesSettings(logger log.Logger, client *http.Client, url *url.URL) *I
...
+					prometheus.BuildFQName(namespace, "indices_settings", "creation_timestamp_seconds"),
+					"index setting creation_date",
+					defaultIndicesTotalFieldsLabels, nil,
+				),
+				Value: func(indexSettings Settings) float64 {
+					val, err := strconv.ParseFloat(indexSettings.IndexInfo.CreationDate, 64)
+					if err != nil {
+						return float64(defaultDateCreation)
+					}
+					return val
```
SuperQ, line 121: I was just looking into this field, it looks like the value is actually expressed as a timestamp in epoch millseconds. So we need to convert this.

```suggestion
					return val/1000.0
```

SuperQ (changes requested):

collector/indices_settings.go
capuche2412: I also precise the unit in the README description

SuperQ (approved):

sysadmind (approved):

PR #830, opened 2023-12-07
Update README.md

xogoodnow:
The following flag has been replaced but it is not mentioned in the main readme file and only in release notes which makes this prone for confusion. --es.cluster_settings has been renamed to --collector.clustersettings


clavinjune:
Hi @xogoodnow can we also change this line https://github.com/prometheus-community/elasticsearch_exporter/blob/master/README.md?plain=1#L91 ?

xogoodnow:
Hi @clavinjune, 
I didn't understand what you said.
Could you please explain what you mean exactly?

clavinjune:
please also remove this one https://github.com/xogoodnow/elasticsearch_exporter/blob/patch-1/README.md?plain=1#L56

clavinjune:
@SuperQ @sysadmind could you please help review this MR :bow:

xogoodnow:
> please also remove this one https://github.com/xogoodnow/elasticsearch_exporter/blob/patch-1/README.md?plain=1#L56

This option "es.indices " is working fine, is there is particular reason on why to omit this line?

xogoodnow:
Hi @clavinjune, on "es.cluster_settings" (L56), I have mentioned that this flag has been deprecated as of v1.6.0.
So IMHO, I believe it would be better for backwards compatibility to keep this flag in the document.

clavinjune:
in 1.6.0, if you use that flag, the exporter won't run. Since the said flag is also marked as breaking changes, I don't think we need to  have that on the readme, wdyt?

I guess we can leave that to the maintainer's judgement

xogoodnow:
> in 1.6.0, if you use that flag, the exporter won't run. Since the said flag is also marked as breaking changes, I don't think we need to have that on the readme, wdyt?
> 
> I guess we can leave that to the maintainer's judgement

Yes I agree.
 also since no other deprecated flag is mentioned in the readme file, for the sake of consistency I deleted the flag .
 
 As an example in V1.4.0 >> Rename elasticsearch_process_cpu_time_seconds_sum to elasticsearch_process_cpu_seconds_total. But there is no mention of "elasticsearch_process_cpu_time_seconds_sum" in the main readme file (Metrics).
 
 In the next release, the flag `--es.snapshots` could be renamed to `--collector.snapshots`. So if that happens, we need to change the readme for this flag as well.
Thank you @clavinjune

sysadmind (approved):
LGTM Thanks!

PR #857, opened 2024-02-05
Update README.md for version 1.7.0

dbalucas:
This commit fixes the Readme where the es.snapshot option is renamed to collector.snapshot in version 1.7.0


README.md
```diff
@@ -58,7 +58,7 @@ elasticsearch_exporter --help
...
-| es.snapshots            | 1.0.4rc1              | If true, query stats for the cluster snapshots.                                                                                                                                                                                                                                                                                                                                       | false |
+| collector.snapshot      | 1.0.4rc1              | If true, query stats for the cluster snapshots. Renamed to collector.snapshots since version 1.7.0                                                                                                                                                                                                                                                                                    | false |
```
sysadmind, line 61: I don't think renamed to collector.snapshot makes sense on the collector.snapshot entry. Maybe this should be `Renamed from es.snapshots in version 1.7.0`

sysadmind (changes requested):

README.md
dbalucas: I agree. Thx

PR #892, opened 2024-05-13
Add missing documentation for ILM metrics

multani:
#513 added the metrics but they were not documented.

I'm not sure about the "cardinality" content ; this is my understanding:

* `elasticsearch_ilm_status` has an `operation_mode` label which can have 3 values so I set cardinality to `3`
* `elasticsearch_ilm_index_status` has [more labels](https://github.com/prometheus-community/elasticsearch_exporter/blob/ea789710cea39aab5a7e96d64943cf7fbb903c4b/collector/ilm_indices.go#L59), in particular `index` which can be unbounded, so I didn't set any cardinality (since it depends on the number of indices in the cluster).


README.md
```diff
@@ -139,6 +140,8 @@ Further Information
 | elasticsearch_filesystem_io_stats_device_write_operations_count      | gauge      | 1           | Count of disk write operations                                                                      |
 | elasticsearch_filesystem_io_stats_device_read_size_kilobytes_sum     | gauge      | 1           | Total kilobytes read from disk                                                                      |
 | elasticsearch_filesystem_io_stats_device_write_size_kilobytes_sum    | gauge      | 1           | Total kilobytes written to disk                                                                     |
+| elasticsearch_ilm_status                                             | gauge      | 3           | Current status of ILM. Status can be `STOPPED`, `RUNNING`, `STOPPING`.                              |
```
sysadmind, line 143: In this case, cardinality is the number of labels, not the number of possible values for the labels. ilm_status should be 1 and ilm_index_status should be 4.

sysadmind (changes requested):
Thanks for the PR, just a small cleanup to be consistent with the rest.

README.md
multani:
```suggestion
| elasticsearch_ilm_status                                             | gauge      | 1           | Current status of ILM. Status can be `STOPPED`, `RUNNING`, `STOPPING`.                              |
```

README.md
```diff
@@ -139,6 +140,8 @@ Further Information
...
+| elasticsearch_ilm_index_status                                       | gauge      |             | Status of ILM policy for index                                                                      |
```
multani, line 144:
```suggestion
| elasticsearch_ilm_index_status                                       | gauge      | 4           | Status of ILM policy for index                                                                      |
```

sysadmind (approved):

PR #917, opened 2024-08-20
CVE-2024-24790: bump go to 1.22.6

### LINKED ISSUE #913: CVE-2024-24790 Related to StdLib
CVE-2024-24790 in elasticsearch-exporter:stdlib
**Description**
The various Is methods (IsPrivate, IsLoopback, etc) did not work as expected for IPv4-mapped IPv6 addresses, returning false for addresses which would return true in their traditional IPv4 forms.

**Severity**:  Critical
**Project**: [------------------/ elasticsearch-exporter](https://gitlab.com/-----------------/elasticsearch-exporter)
**Tool**: Container Scanning
**Scanner**: Trivy
**Location**
**Image**: elasticsearch-exporter:latest
**Namespace**: Unknown
**Links**
http://www.openwall.com/lists/oss-security/2024/06/04/1
https://access.redhat.com/security/cve/CVE-2024-24790
https://go.dev/cl/590316
https://go.dev/issue/67680
https://groups.google.com/g/golang-announce/c/XbxouI9gY7k/m/TuoGEhxIEwAJ
https://nvd.nist.gov/vuln/detail/CVE-2024-24790
https://pkg.go.dev/vuln/GO-2024-2887
https://www.cve.org/CVERecord?id=CVE-2024-24790
**Identifiers**
[CVE-2024-24790](http://www.openwall.com/lists/oss-security/2024/06/04/1)
**Solution**
Upgrade stdlib to 1.21.11, 1.22.4

**Evidence**
Vendor Status
**fixed
Vulnerable Package**
stdlib:1.21.5

sdx-jkataja:
Fixes vulnerability [CVE-2024-24790](https://nvd.nist.gov/vuln/detail/CVE-2024-24790) by updating Go to 1.22.6

Fixes #913

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ executors:
   # This must match .promu.yml.
   golang:
     docker:
-      - image: cimg/go:1.21
+      - image: cimg/go:1.22.6
 jobs:
   test:
     executor: golang
--- a/.promu.yml
+++ b/.promu.yml
@@ -1,6 +1,6 @@
 go:
     # This must match .circle/config.yml.
-    version: 1.21
+    version: 1.22
 repository:
     path: github.com/prometheus-community/elasticsearch_exporter
 build:


.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ executors:
   # This must match .promu.yml.
   golang:
     docker:
-      - image: cimg/go:1.21
+      - image: cimg/go:1.22.6
```
sysadmind, line 9: This needs to be changed in  `.promu.yml` as well although there it should just be `1.22`. 
See here: https://github.com/prometheus/prometheus/blob/main/.promu.yml

sysadmind (changes requested):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/prometheus-community/elasticsearch_exporter
 
-go 1.20
+go 1.22.6
```
sysadmind, line 3: I think with a version this new, the toolchain declaration is required.

.circleci/config.yml
sdx-jkataja: Commit fe96c8f1cfe52b7bb644ebed66ad3b8095555016

go.mod
sdx-jkataja: Please advise on what's needed for the toolchain declaration

sdx-jkataja: Do you mean the following declaration?
```
go 1.22.6

toolchain go1.22.6
```
Running `go mod tidy` removes  the `toolchain` line, so I assume it won't be required.

sysadmind: I think it depends on your local go version. After a second thought, we don't need to change this file at all. This is only a minimum go version and we don't need any new language features so we can leave this as 1.20

sdx-jkataja: I'm go newbie so I had to check, and it appears you're correct. I have go 1.22.5 and when I built the code in master branch and then scan with Syft, it reports that the image contains go standard library 1.22.5. 

Should I remove the change in `go.mod` and rely on CI files to use the updated go version?

sdx-jkataja: Done, I removed the commit

sysadmind (approved):

mohhai1:
@sysadmind , kindly, when do we expect w release/tag with the fix?

sysadmind:
https://github.com/prometheus-community/elasticsearch_exporter/releases/tag/v1.8.0

PR #920, opened 2024-08-30
support for scraping multiple elasticsearch instance

lework:
Pass the es url via the url parameter.

```
export ES_TEST_USERNAME elastic
export ES_TEST_PASSWORD 123456

http://localhost:9114/probe?target=http://test-elasticsearch-01:9200&auth_module=TEST
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,9 @@ elasticsearch_exporter:
     - "127.0.0.1:9114:9114"
 ```
 
+- /metrics endpoint is used to scrape the elasticsearch-exporter.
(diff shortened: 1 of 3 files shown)


sysadmind (changes requested):
The code, as it is currently, completely removes the metrics collection for the `/metrics` endpoint. We can't break the single instance functionality when adding the multi target support. 

See this PR for a reference where I added multi target support to the postgres_exporter https://github.com/prometheus-community/postgres_exporter/pull/618

PR #928, opened 2024-09-07
Refactor tests for slm collector

sysadmind:
- Remove up, totalScrapes, and jsonParseFailures metrics. They are not useful.
- Move fixtures to individual files
- Base tests on the metric output for better testing the expected output instead of the internals.


SuperQ:
I think the reason we have the `up` metric is that the `Collect()` doesn't have any error return. So we can't pass back a failed scrape signal to the output. Without this, the exporter will silently fail, which is not good for the end user.

I think we need a bit more refactor here, so that `error` is returned and we fail the scrape with a 5xx error instead of returning partial data.

sysadmind:
I agree that knowing the collector as a whole or an individual collector was successful is useful. This is provided by the collector/Collector going forward: https://github.com/prometheus-community/elasticsearch_exporter/blob/master/collector/collector.go#L57

This particular change is just to update the testing before refactoring the slm collector to work with the Collector interface which will reintroduce the ability to determine if the collector scrape was successful. There have been a handful of similar PRs in this light. See https://github.com/prometheus-community/elasticsearch_exporter/pull/813

The follow up PR will will refactor the code to use the Collector interface. i.e. https://github.com/prometheus-community/elasticsearch_exporter/pull/789

I have been breaking up the refactor from the testing changes to make PRs easier to review and to make sure the refactors don't meaningfully change the metrics. If you would prefer a different strategy that is easier to review or more to your preference, let me know.

SuperQ (approved):

PR #933, opened 2024-09-16
added external refresh stats

TomaBere:


collector/nodes.go
```diff
@@ -630,6 +629,30 @@ func NewNodes(logger log.Logger, client *http.Client, url *url.URL, all bool, no
...
+				),
+				Value: func(node NodeStatsNodeResponse) float64 {
+					return float64(node.Indices.Refresh.ExternalTotal)
+				},
+				Labels: defaultNodeLabelValues,
+			},
+			{
+				Type: prometheus.CounterValue,
+				Desc: prometheus.NewDesc(
+					prometheus.BuildFQName(namespace, "indices_refresh", "external_total_time_in_millis"),
```
sysadmind, line 647: This is actually seconds in the name. The value is being divided to get back to seconds and prometheus best practices are to store this data in seconds.

collector/nodes.go
```diff
@@ -16,14 +16,13 @@ package collector
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/go-kit/log"
```
sysadmind, line 19: These imports should not be moved. They're formatted with goimports

collector/indices_response.go
```diff
@@ -145,9 +145,11 @@ type IndexStatsIndexMergesResponse struct {
 
 // IndexStatsIndexRefreshResponse defines index stats index refresh information structure
 type IndexStatsIndexRefreshResponse struct {
-	Total             int64 `json:"total"`
-	TotalTimeInMillis int64 `json:"total_time_in_millis"`
-	Listeners         int64 `json:"listeners"`
+	Total                     int64 `json:"total"`
+	TotalTimeInMillis         int64 `json:"total_time_in_millis"`
+	ExternalTotal             int64 `json:"external_total"`
+	ExternalTotalTimeInMillis int64 `json:"external_total_time_in_millis"`
```
sysadmind, line 151: It looks like the only additional field we're using is ExternalTotalTimeInMillis. I think we should only add that field for now.

sysadmind (changes requested):

collector/indices_response.go
TomaBere: Hey, 
added ExternalTotal & ExternalTotalTimeInMillis.

collector/nodes.go
TomaBere: cool, changed to external_total_time_in_seconds .

sysadmind (approved):

PR #959, opened 2024-11-10
Add github actions workflow for testing PRs

sysadmind:
This may need some more work because it's not always easy to test actions locally. After this runs well for all PRs and releases, I will remove the circleCI workflows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,3 +17,79 @@ jobs:
       - uses: prometheus/promci@52c7012f5f0070d7281b8db4a119e21341d43c91 # v0.4.5
       - uses: ./.github/promci/actions/setup_environment
       - run: make GO_ONLY=1 SKIP_GOLANGCI_LINT=1
+
+  build:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,102 @@
+---
+name: CI
+on:
+  pull_request:
+  push:
```
SuperQ, line 5: The only downside to this is that this doubles the CI runs for contributors. Since this matches Both the PR and the Push.

Maybe we can figure out a filter so this doesn't happen.

SuperQ (approved):
This seems fine. I wish we could make our GH actions a little less verbose.

.github/workflows/ci.yml
sysadmind: I copied this from the CI job for the prometheus repo https://github.com/prometheus/prometheus/blob/main/.github/workflows/ci.yml#L5. Does that job run twice? If this runs twice here when someone opens a PR, I'm happy to filter to the master/main branches.

PR #1000, opened 2025-02-19
Fix github actions docker image org

sysadmind:
Actions are currently failing to publish the images because the image names are incorrect. I'm not positive if there are more changes necessary yet, but this should correct the organization name.

See https://github.com/prometheus-community/elasticsearch_exporter/actions/runs/13350556606/job/37286525153


jan--f (approved):
I think this is what's needed to publish images.

PR #942, opened 2024-10-14
chore!: adopt log/slog, drop go-kit/log

tjhop:
The bulk of this change set was automated by the following script which
is being used to aid in converting the various exporters/projects to use
slog:

https://gist.github.com/tjhop/49f96fb7ebbe55b12deee0b0312d8434

In addition to the parts that were straightforward conversions, this
also:
- refactors much of the logging config to adopt slog
    - removed custom `logger.go` setup for go-kit
    - adopt promslog/flag and use that to handle parsing log
      level/format flags
    - for consistent behavior, keep log output flag to allow toggle
      stdout/stderr for output
    - adopt promslog for logger setup
- enables sloglint in golangci-lint config
- drops go-kit/log exclusions from lint config
- tidies mods to drop go-kit/log and go-logfmt/logfmt deps

PR #1003, opened 2025-02-26
Release v1.9.0

SuperQ:
BREAKING CHANGES:

The flag `--es.slm` has been renamed to `--collector.slm`.

The logging system has been replaced with log/slog from the stdlib. This change is being made across the prometheus ecosystem. The logging output has changed, but the messages and levels remain the same. The `ts` label for the timestamp has bewen replaced with `time`, the accuracy is less, and the timezone is not forced to UTC. The `caller` field has been replaced by the `source` field, which now includes the full path to the source file. The `level` field now exposes the log level in capital letters.

* [CHANGE] Rename --es.slm to --collector.slm #932
* [CHANGE] Replace logging system #942
* [ENHANCEMENT] Add external refresh stats #933


dswarbrick (approved):

sysadmind:
This would be the first release that is run by GitHub Actions, so we should pay extra attention. I think we need to merge #1000 first. I'm not sure but maybe we should also disable the CircleCI workflow before merging this as well.

PR #1014, opened 2025-03-05
Refactor indices collector

sysadmind:
- Move metric DESC to vars to aid in unused linter checks
- Clean up HTTP requests and JSON handling

Will refactor to use the Collector interface in a future change

### DIFF
--- a/collector/indices.go
+++ b/collector/indices.go
@@ -27,376 +27,378 @@ import (
 )
 
 var (
+	indicesLabels = []string{"index", "cluster"}
+
 	indicesDocsPrimary = prometheus.NewDesc(
 		prometheus.BuildFQName(namespace, "indices", "docs_primary"),
 		"Count of documents with only primary shards",
-		[]string{"index", "cluster"}, nil,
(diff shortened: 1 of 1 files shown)


collector/indices.go
```diff
@@ -14,44 +14,421 @@
...
-type labels struct {
-	keys   func(...string) []string
-	values func(*clusterinfo.Response, ...string) []string
-}
+var (
+	indicesDocsPrimary = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, "indices", "docs_primary"),
+		"Count of documents with only primary shards",
+		[]string{"index", "cluster"}, nil,
+	)
```
SuperQ, line 34: Let's make these labels a var so there's less chance of copy-pasta mistakes.

```suggestion
	indicesLabels = []string{"index", "cluster"}
	indicesDocsPrimary = prometheus.NewDesc(
		prometheus.BuildFQName(namespace, "indices", "docs_primary"),
		"Count of documents with only primary shards",
		indicesLabels, nil,
	)
```

PR #1016, opened 2025-03-10
Remove unused golangci-lint config

sysadmind:
This config isn't needed because we're the code doesn't fail the linter check. The config is however preventing golangci-lint from being upgraded because this does not pass the schema validation.


SuperQ (approved):
Thanks

PR #1027, opened 2025-04-02
Remove copyright dates

sysadmind:
This should help when PRs are open through a year change. These dates aren't necessary and the CNCF does not recommend keeping them.

https://github.com/cncf/foundation/blob/main/copyright-notices.md


sysadmind:
bump

SuperQ:
Seems like something we should decide on project-wide. I don't have any objection, but it would be good to make sure everyone is in agreement.

SuperQ (approved):
LGTM!

PR #1033, opened 2025-04-23
Add pprof for profiling

sysadmind:
We have received reports of excessive memory usage for the exporter. Pprof provides a mechanism for users to dump the heap which should provide useful information about what memory has been allocated.

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	_ "net/http/pprof" // Enable pprof metrics
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	_ "net/http/pprof" // Enable pprof metrics
```
SuperQ, line 20:
```suggestion
	_ "net/http/pprof"
```

SuperQ (approved):
Minor nit, otherwise LGTM.

PR #1034, opened 2025-04-23
Update golangci-lint config for v2

sysadmind:
This brings the common files in line with other prometheus repos. Most of these settings are coppied from prometheus/prometheus. Where it was a small change, I have enabled the new linte rs and made the fixes. The remaining linters that are commented out deserve to be enabled and the fixes applied as independent changes.


SuperQ:
You should be able to run `golangci-lint migrate` to convert the v1 config to v2.

SuperQ (approved):

PR #1049, opened 2025-06-11
Fix memleak for create a TimeTicker without stop

### LINKED ISSUE #1048: Serious memory leak issue
I had a very serious memory leak when using elasticsearch_exporter to collect metrics, which was causing my app to OOM every few hours in production.
Through the analysis of the pprof tool, the memory usage is relatively large as follows:
Image
After analyzing the code, it is found that in the `Run` function of the `clusterinfo` module, a `TimeTicker` is defined but it is not stopped.
```go
go func(ctx context.Context) {
		if r.interval <= 0 {
			r.logger.Info("no periodic cluster info label update requested")
			return
		}
		ticker := time.NewTicker(r.interval)
		for {
			select {
			case <-ctx.Done():
				r.logger.Info(
					"context cancelled, exiting cluster info trigger loop",
					"err", ctx.Err(),
				)
				return
			case <-ticker.C:
				r.logger.Debug("triggering periodic update")
				r.sync <- struct{}{}
			}
		}
	}(ctx)
```
When this function is requested on a regular basis, a large number of timers are generated that are not released, causing a serious memory leak.

YenchangChan:
Fixed #1048   memleak for create a TimeTicker without stop

### DIFF
--- a/pkg/clusterinfo/clusterinfo.go
+++ b/pkg/clusterinfo/clusterinfo.go
@@ -218,6 +218,7 @@ func (r *Retriever) Run(ctx context.Context) error {
 			return
 		}
 		ticker := time.NewTicker(r.interval)
+		defer ticker.Stop()
 		for {
 			select {
 			case <-ctx.Done():


pkg/clusterinfo/clusterinfo.go
```diff
@@ -225,6 +225,7 @@ func (r *Retriever) Run(ctx context.Context) error {
 					"context cancelled, exiting cluster info trigger loop",
 					"err", ctx.Err(),
 				)
+				ticker.Stop()
```
sysadmind, line 228: I think this should be a `defer ticker.Stop()` just after creation on line 220 instead. That way we can't accidentally have this same problem in the future.

sysadmind (changes requested):
This commit needs the DCO to pass checks. You should be able to `git commit --ammend --signoff`. 

While I agree that this change is positive, I don't think this will solve a memory problem. The Run function is only ever called once from main.go, so I don't see how we could make multiple timers. I suspect something else is going on here. Really glad to see that pprof is helping here!

pkg/clusterinfo/clusterinfo.go
YenchangChan: Indeed, this is not easy to occur. But our usage scenarios are a little different: We developed a collector for collecting prometheus metrics by imitating the retrieval module of prometheus and integrated some commonly used exporters, including elasticsearch_exporter, which is used as a libaray. And it is called by the main program at regular intervals, which is why such a serious memory problem occurred.

BTW, the code has been adjusted according to your suggestions. Please review the PR. Thank you.

sysadmind (approved):

PR #1050, opened 2025-06-12
Don't log collector errors

SuperQ:
Avoid logging collector failures at Error level to avoid log spam.


collector/collector.go
```diff
@@ -185,7 +185,7 @@ func execute(ctx context.Context, name string, c Collector, ch chan<- prometheus
 		if IsNoDataError(err) {
 			logger.Debug("collector returned no data", "name", name, "duration_seconds", duration.Seconds(), "err", err)
 		} else {
-			logger.Error("collector failed", "name", name, "duration_seconds", duration.Seconds(), "err", err)
+			logger.Debug("collector failed", "name", name, "duration_seconds", duration.Seconds(), "err", err)
```
sysadmind, line 188: I think Error is the right level here. It really is an error.

sysadmind (approved):

PR #1051, opened 2025-06-12
Do not log input URL string

sysadmind:
The error itself already has the sanitized URL so there is no need to log the sensitive input string.


sysadmind:
bump

SuperQ (approved):

PR #1052, opened 2025-06-18
Create a new cluster info retriever

sysadmind:
This is an alternative implementation of the previous retriever in pkg/clusterinfo. The previous implementation required registering collectors and receiving updates over a channel. This InfoProvider instead provides a public method to get the cluster info with a cache to reduce network calls. The Collector.Update signature was also updated to take an UpdateContext. This UpdateContext can provide with an easy way to extend the context available during the Update calls on Collectors.

### DIFF
--- a/cluster/provider.go
+++ b/cluster/provider.go
@@ -1,0 +1,117 @@
+// Copyright The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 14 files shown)


sysadmind:
bump @SuperQ

SuperQ (commented):
Hmm, I'm not so sure about the cache implementation.

Is it possible that this could leak entries in the cache unbounded?

sysadmind:
@SuperQ can you elaborate?

SuperQ (approved):
Yea, I guess I was misunderstanding the cache implementation.

PR #1123, opened 2026-01-15
Fix: Preserve Elasticsearch URL path in stats requests

Robert-Steiner:
# Description

When the Elasticsearch URL contains a path (e.g., `http://example.com/elasticsearch/`), the path is incorrectly stripped when making `_all/_stats` requests, resulting in requests to `http://example.com/_all/_stats` instead of `http://example.com/elasticsearch/_all/_stats`.

The regression was introduced in v1.10.0 when the URL resolution logic was changed from using [`path.Join`](https://github.com/prometheus-community/elasticsearch_exporter/blob/v1.9.0/collector/indices.go#L1116) to `url.ResolveReference`.

```go
package main

import (
    "fmt"
    "net/url"
)

func main() {
    base, _ := url.Parse("http://example.com/elasticsearch/")
    fmt.Println(base.ResolveReference(&url.URL{Path: "/_all/_stats"}))
    fmt.Println(base.ResolveReference(&url.URL{Path: "_all/_stats"}))
}
```

Output

```
http://example.com/_all/_stats
http://example.com/elasticsearch/_all/_stats
```

This PR fixes the URL path handling to ensure custom paths in the Elasticsearch URL are properly preserved when constructing API endpoint requests.

### DIFF
--- a/collector/indices.go
+++ b/collector/indices.go
@@ -574,7 +574,7 @@ func (i *Indices) Describe(ch chan<- *prometheus.Desc) {
 func (i *Indices) fetchAndDecodeIndexStats(ctx context.Context) (indexStatsResponse, error) {
 	var isr indexStatsResponse
 
-	u := i.url.ResolveReference(&url.URL{Path: "/_all/_stats"})
+	u := i.url.ResolveReference(&url.URL{Path: "_all/_stats"})
 	q := u.Query()
 	q.Set("ignore_unavailable", "true")
 	if i.shards {


collector/indices.go
```diff
@@ -574,7 +574,7 @@ func (i *Indices) Describe(ch chan<- *prometheus.Desc) {
 func (i *Indices) fetchAndDecodeIndexStats(ctx context.Context) (indexStatsResponse, error) {
 	var isr indexStatsResponse
 
-	u := i.url.ResolveReference(&url.URL{Path: "/_all/_stats"})
+	u := i.url.ResolveReference(&url.URL{Path: "_all/_stats"})
```
ArthurSens, line 577: Maybe we should be using `i.url.JoinPath` here?

ArthurSens (changes requested):
Hi @Robert-Steiner, sorry for the delay here. Looks like the PR needs a rebase now.

Also, could you add a test to prevent us from re-introducing this bug again in the future?

collector/indices.go
sysadmind: Based on my research, I think JoinPath should become our standard. It covers the prefix needs we have.

PR #1144, opened 2026-03-05
fix: optimize datastream log and resolve thread leak

0x0034:
1.  optimize data_stream log print 
2. resolve shards collector thread leak


ArthurSens:
Thanks for working on this. The data-stream logging change looks good directionally, but the goroutine fix is unsafe as written.

Shards is a long-lived cluster-info consumer in single-target mode. Closing its channel after two minutes leaves the retriever holding that channel, so its next update—five minutes by default—will panic with `send on closed channel`.

We have a PR open at #1226 with good results, but we're still deciding whether to remove that part of the codebase.

With that said, I'm closing this one, thanks again!

PR #1157, opened 2026-04-15
feat: add optional node label to elasticsearch_indices_aliases metric

### LINKED ISSUE #1156: [FEATURE] Add optional node label to elasticsearch_indices_aliases metric
### Feature Request

The `elasticsearch_indices_aliases` metric currently exposes labels `{index, alias, cluster}`.
There is no way to correlate an alias with the Elasticsearch nodes hosting its underlying index.

### Why?

- Allows PromQL queries to answer "which nodes serve traffic for alias X?" for capacity
  planning, topology-aware routing, and debugging.
- Enables joining alias metrics with node-level metrics to identify hotspots and indexing
  throttle issues without external label relabeling.
- Similar to how `elasticsearch_nodes_roles` already exposes a `node` label (#1111).

### Proposed Implementation

Add an opt-in `--es.aliases.node-labels` flag (default `false`) that appends a `node` label
to the `elasticsearch_indices_aliases` metric using shard routing data.

**When disabled (default):** metric is unchanged — `{index, alias, cluster}`. Zero breakage.

**When enabled:** metric gains a 4th `node` label — one series per `(index, alias, node)`.
When `--es.shards` is disabled, an extra lightweight call to
`/_all/_stats?level=shards&metric=docs` fetches shard-to-node placement.

**Cardinality note:** the multiplier is `unique_nodes_per_index` (typically 3-10), not total
shards. For a cluster with 100 indices × 1 alias × 3 nodes = 300 series vs 100 without
the flag. The flag defaults to `false` so there is no impact unless explicitly enabled.

pincher95:
## Summary

Adds an opt-in `--es.aliases.node-labels` flag that appends a `node` label to the
`elasticsearch_indices_aliases` metric, showing which nodes host shards for each
aliased index.

- When disabled (default): metric keeps the original 3 labels `{index, alias, cluster}` — fully backward compatible
- When enabled: emits one series per `(index, alias, node)` with 4 labels `{index, alias, node, cluster}`

Resolves #1156

## Changes

**`collector/indices.go`**
- Added `indicesAliasesWithNode` descriptor with 4 labels alongside the original 3-label `indicesAliases`
- Added `aliasNodeLabels` and `aliasesDesc` fields to `Indices` struct; constructor selects descriptor based on flag
- Added `populateShardRouting()` to fetch shard placement when `--es.shards` is off
- Added `collectAliasNodes()` helper to extract sorted unique node IDs from shard data
- `Collect()` branches on `aliasNodeLabels`: legacy path emits 3 labels, node-aware path emits per-node series

**`collector/indices_test.go`**
- Existing test cases unchanged (verify backward-compatible 3-label output)
- Added `TestIndicesAliasNodeLabels` validating node labels are populated from shard routing fixtures

**`main.go`**
- Added `--es.aliases.node-labels` flag (default `false`)
- Passed to `NewIndices()` in both single-target and `/probe` paths

**`README.md`**
- Documented new flag in CLI table


collector/indices.go
```diff
@@ -640,11 +658,52 @@ func (i *Indices) fetchAndDecodeIndexStats(ctx context.Context) (indexStatsRespo
...
+				i.logger.Warn("failed to fetch shard-level routing information for aliases", "err", err)
+			}
+		}
 	}
 
 	return isr, nil
 }
 
+func (i *Indices) populateShardRouting(ctx context.Context, isr *indexStatsResponse) error {
+	u := i.url.ResolveReference(&url.URL{Path: "/_all/_stats"})
```
sysadmind, line 717: This won't work for clusters that have a path prefix because the reference has a leading slash.

sysadmind:
Isn't this already possible by joining labels in Prometheus? There is already an `elasticsearch_indices_aliases` which would allow you to know which index an alias is associated with. And from that, you could know the node that an index is associated with right? I think this is already possible without adding more functionality.

PR #1159, opened 2026-04-15
perf(indices): stream /_all/_stats decode to cut peak memory ~10x

### LINKED ISSUE #1158: [PERF] Replace io.ReadAll + json.Unmarshal with streaming json.NewDecoder
### Performance Issue

The exporter currently buffers entire HTTP response bodies into memory (`io.ReadAll` + `json.Unmarshal`) for every Elasticsearch API call. In multi-cluster environments using the `/probe` endpoint, this causes:

- **Memory spikes:** Each scrape allocates the full response body as a `[]byte` buffer before parsing. Node stats responses from large clusters (50+ nodes) can be 5-20 MB. With concurrent `/probe` scrapes across N clusters, peak memory = `N × response_size × 2` (the `[]byte` buffer + the parsed Go structs coexist until GC collects the buffer).
- **GC pressure:** Short-lived multi-MB allocations cause frequent GC cycles, increasing tail latency.

### Proposed Fix

Replace `io.ReadAll` + `json.Unmarshal` with `json.NewDecoder(resp.Body).Decode()` across all collectors. This streams JSON directly from the HTTP response into Go structs without an intermediate buffer, halving per-request memory allocations.

This is a pure refactor with no behavioral change — existing tests verify identical metric output.

pincher95:
## Summary

Two related changes to how collectors fetch and decode Elasticsearch responses:

1. **perf(indices)** — Stream-decode the `/_all/_stats` response one index at a
   time instead of buffering the whole body and unmarshaling it into a map of
   every index. The collector only iterates indices and emits per-index metrics,
   so the full map never needs to be in memory. On large clusters this response
   is tens of MB, and holding the decoded map live for the emit loop dominates
   peak RSS — multiplied under concurrent `/probe` scrapes.

2. **refactor** — Add a `getAndDecodeURL` helper (and lower-level `fetchURL`)
   consolidating the duplicated GET + read + decode + body-close boilerplate, and
   fix the `resp.Body.Close()` error handling that shadowed return values.

This supersedes the original idea in this PR (swapping `io.ReadAll`+`json.Unmarshal`
for `json.NewDecoder().Decode()`), which benchmarking showed does not reduce memory.

## Why not json.NewDecoder().Decode()?

`json.NewDecoder(resp.Body).Decode()` buffers the entire body in its internal
buffer (grown by doubling) and then unmarshals, so for single-shot decoding it
allocates the same data plus the `Decoder` — equal or more bytes/op than
`io.ReadAll`+`json.Unmarshal` across payloads from 13 KB to 6.6 MB
(`collector/decode_bench_test.go`, `collector/decode_realworld_test.go`). So
`getAndDecodeURL` keeps `io.ReadAll`+`json.Unmarshal`. The real win is not the
decode call — it is not materializing the whole response.

## Benchmarks (synthetic, BenchmarkIndexStatsDecode)

| payload | approach | bytes/op | allocs/op | throughput |
|---|---|---|---|---|
| 672 KB (100 indices) | readall+unmarshal+map | 1.65 MB | 548 | 175 MB/s |
| 672 KB | streaming | **0.18 MB (-89%)** | 1043 | 177 MB/s |
| 6.6 MB (1000 indices) | readall+unmarshal+map | 17.1 MB | 5065 | 179 MB/s |
| 6.6 MB | streaming | **1.6 MB (-91%)** | 10043 | 178 MB/s |

The committed suite uses a synthetic `_all/_stats` generator (`buildIndicesPayload`)
so it needs no external cluster.

## Production validation

The same comparison run against real `/_all/_stats` responses captured from four
production clusters of varying size and region (identifiers masked), decoded into
the production structs:

| cluster | response | peak heap (map → stream) | bytes/op (map → stream) | throughput |
|---|---|---|---|---|
| A | 10 MB | 25 → 2.7 MB (**9.4x**) | 26.2 → 2.8 MB (**-89%**) | ~170 MB/s |
| B | 35 MB | 84 → 10.3 MB (**8.1x**) | 88.2 → 10.8 MB (**-88%**) | ~163 MB/s |
| C | 39 MB | 89 → 9.9 MB (**8.9x**) | 93.0 → 10.4 MB (**-89%**) | ~163 MB/s |
| D | 52 MB | 125 → 13.1 MB (**9.5x**) | 131 → 13.8 MB (**-89%**) | ~161 MB/s |

Peak live heap drops ~8–9.5x and cumulative allocation ~88–89% across all four,
throughput unchanged. Allocation *count* rises ~1.5x (each index is decoded
separately) but each is tiny and short-lived; total allocated bytes — what drives
GC — is what falls. (These ran against locally captured responses and are not part
of the committed test suite.)

## Correctness

Per-index metric output is identical: the existing `TestIndices` `CollectAndCompare`
golden tests pass unchanged, plus a new `TestStreamIndexStatsEquivalence` asserts
the streamed decode equals `json.Unmarshal` (including `level=shards`).

## Note

On a mid-stream decode error (e.g. a truncated body) some index metrics may
already have been emitted, vs the previous all-or-nothing buffered decode. This
only affects the error path and is the inherent trade-off of streaming.

Resolves #1158

### DIFF
--- a/collector/cluster_health.go
+++ b/collector/cluster_health.go
@@ -16,7 +16,6 @@ package collector
 import (
 	"encoding/json"
 	"fmt"
-	"io"
 	"log/slog"
 	"net/http"
 	"net/url"
@@ -235,12 +234,7 @@ func (c *ClusterHealth) fetchAndDecodeClusterHealth() (clusterHealthResponse, er
 		return chr, fmt.Errorf("HTTP Request failed with code %d", res.StatusCode)
 	}
 
-	bts, err := io.ReadAll(res.Body)
-	if err != nil {
-		return chr, err
-	}
-
-	if err := json.Unmarshal(bts, &chr); err != nil {
+	if err := json.NewDecoder(res.Body).Decode(&chr); err != nil {
 		return chr, err
 	}
 
--- a/collector/cluster_info.go
+++ b/collector/cluster_info.go
@@ -16,7 +16,6 @@ package collector
 import (
 	"context"
 	"encoding/json"
-	"io"
 	"log/slog"
 	"net/http"
 	"net/url"
@@ -78,18 +77,19 @@ type VersionInfo struct {
 }
 
 func (c *ClusterInfoCollector) Update(_ context.Context, ch chan<- prometheus.Metric) error {
+	var info ClusterInfoResponse
+
 	resp, err := c.hc.Get(c.u.String())
 	if err != nil {
 		return err
 	}
-	defer resp.Body.Close()
-	b, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return err
-	}
-	var info ClusterInfoResponse
-	err = json.Unmarshal(b, &info)
-	if err != nil {
+	defer func() {
+		if cerr := resp.Body.Close(); cerr != nil {
+			c.logger.Warn("failed to close response body", "err", cerr)
+		}
+	}()
+
+	if err := json.NewDecoder(resp.Body).Decode(&info); err != nil {
 		return err
 	}
 
--- a/collector/cluster_info.go
+++ b/collector/cluster_info.go
@@ -78,18 +78,24 @@ type VersionInfo struct {
 }
 
 func (c *ClusterInfoCollector) Update(_ context.Context, uc UpdateContext, ch chan<- prometheus.Metric) error {
+	var info ClusterInfoResponse
+
 	resp, err := c.hc.Get(c.u.String())
 	if err != nil {
 		return err
 	}
-	defer resp.Body.Close()
-	b, err := io.ReadAll(resp.Body)
+	defer func() {
+		if cerr := resp.Body.Close(); cerr != nil {
+			c.logger.Warn("failed to close response body", "err", cerr)
+		}
+	}()
+
+	bts, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return err
 	}
-	var info ClusterInfoResponse
-	err = json.Unmarshal(b, &info)
-	if err != nil {
+
+	if err := json.Unmarshal(bts, &info); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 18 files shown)


sysadmind:
Can you add some benchmarks? I'm not sure how it would reduce the memory footprint based on my reading of the json package in the std library. If you're seeing empirical evidence of such, I would be more suspicious of having the io.ReadAll and json.Unmarshal in the same function and the compiler doing something more intelligent with that less complex structure.

pincher95:
You were right to be skeptical. I added benchmarks (`collector/decode_bench_test.go`, `decode_realworld_test.go`) and `json.NewDecoder().Decode()` does not reduce memory for single-shot decoding — it buffers the whole body internally and unmarshals from there, allocating the same data plus the `Decoder`, so it's equal-or-worse than `io.ReadAll`+`json.Unmarshal` (~31% more bytes, ~24% more GC cycles in a 64-way concurrent decode). I dropped that change; `getAndDecodeURL` keeps `io.ReadAll`+`json.Unmarshal`.

The actual memory cost is that the indices collector buffers the entire `/_all/_stats` response and builds a map of every index, holding it live for the whole emit loop — tens of MB on large clusters, multiplied under concurrent `/probe`. I reworked the PR to stream-decode that response one index at a time and emit immediately (`BenchmarkIndexStatsDecode`): ~90% fewer bytes/op and ~10x lower peak heap (89 MB → 9 MB on a 39 MB payload), throughput unchanged. I also validated against four of our production clusters (10–52 MB `_all/_stats` responses) — same ~9x peak-heap / ~89% allocation reduction, throughput unchanged.

Output is verified identical by the existing `CollectAndCompare` golden tests plus a new equivalence test. Retitled and updated the description. Thanks for pushing back on the original claim.

sysadmind:
Can you rebase? Sorry I didn't realize that another PR would conflict but it should be a clean rebase. Otherwise, this looks amazing and is very thorough. Thanks for the leg work!

pincher95:
@sysadmind rebased

sysadmind (approved):

PR #1168, opened 2026-05-06
fix: Attempt HTTP/2 for HTTP client

### LINKED ISSUE #1167: [BUG] OpenSearch requests fail with 401 when using certificate authentication and `transport-reactor-netty4`
`elasticsearch_exporter` requests against OpenSearch fail with `401 Unauthorized` when OpenSearch is configured with certificate authentication and the `transport-reactor-netty4` HTTP transport.

The issue appears to happen because the client uses HTTP/1.1 by default, and with OpenSearch's `transport-reactor-netty4` / `reactor-netty4-secure` transport the certificate authorization information does not seem to be transmitted or handled correctly over HTTP/1.1.

Pigueiras:
Set ForceAttemptHTTP2 on the HTTP transports used by the main client and the probe client so certificate-authenticated OpenSearch endpoints can negotiate HTTP/2 when it is available.

This keeps HTTP/1.1 fallback while avoiding 401 responses with transport-reactor-netty4, where client certificate authorization can fail over the default HTTP/1.1 path.

Fixes #1167

### DIFF
--- a/main.go
+++ b/main.go
@@ -170,8 +170,9 @@ func main() {
 		var httpTransport http.RoundTripper
 
 		httpTransport = &http.Transport{
-			TLSClientConfig: tlsConfig,
-			Proxy:           http.ProxyFromEnvironment,
+			TLSClientConfig:   tlsConfig,
+			Proxy:             http.ProxyFromEnvironment,
+			ForceAttemptHTTP2: true,
 		}
 
 		esAPIKey := os.Getenv("ES_API_KEY")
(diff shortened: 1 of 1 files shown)


Pigueiras:
In principle it shouldn't affect anything important because it fallback to HTTP/1.1 if the server is not talking HTTP2, but if you feel more comfortable I can create a flag to explicitly enable this option.

sysadmind (approved):

sysadmind:
Looks like we do something similar in client_golang, so I think this is probably fine to do here too. 

https://github.com/prometheus/client_golang/blob/main/api/client.go#L37

PR #1179, opened 2026-06-03
collector: add nodes_hot_threads collector for CPU usage metrics

### LINKED ISSUE #949: elasticsearch_hot_thread_cpu_usage metric is not avail
i have installed es exporter for getting es data and i am trying to find elasticsearch_hot_thread_cpu_usage metric but not able to find such metric also.

ulascansenturk:
## Summary

Adds a new opt-in collector that exposes hot thread CPU usage as a Prometheus metric.

**New metric:** `elasticsearch_node_hot_thread_cpu_usage_ratio{cluster, node, thread}`  
**New flag:** `--collector.nodes_hot_threads` (disabled by default — the API samples over an interval, making it more expensive than stats endpoints)

The collector calls `GET /_nodes/hot_threads?type=cpu` and parses the text response. Thread labels are derived from the thread pool name and thread ID (e.g. `search[T#3]`, `bulk[T#1]`).

Both response formats are supported:
- Old: `75.0% (375ms out of 500ms) cpu usage by thread '...'`
- New (8.x): `75.0% [cpu=75.0%, idle=25.0%] (375ms out of 500ms) cpu usage by thread '...'`

Closes #949


sysadmind (approved):

sysadmind:
Overall this is good, just need to switch the regex dependency to match the prometheus ecosystem standard `github.com/grafana/regexp`

PR #1181, opened 2026-06-04
Refactor indices_settings collector

sysadmind:
- Move metric Desc to package vars
- Use new COllector interface
- Collapse response struct into the indices_settings.go file
- Consolidate HTTP code into the Update() function

The tests haven't changed - this is a refactor only.


ArthurSens (commented):
I believe README also needs to be updated with the new flag. And what is the plan with the current `es.indices` flag?

PR #1193, opened 2026-06-17
fix: Use logger.Debug instead of Printf for data stream stats

adjurk:
The log fires once per data stream per scrape. With a 15s scrape interval and 20 data streams that's ~80 lines/min written directly to stdout, bypassing the structured logger entirely.

- collector/data_stream.go contained a leftover `fmt.Printf("Metric: %+v", dataStream)` debug statement that printed to stdout on every scrape for every data stream with no way to suppress it
- Replaced with `ds.logger.Debug()` using structured key/value fields, consistent with the pattern used in indices.go and shards.go
- The output is now silent at default log levels and only appears when running with `--log.level=debug`

### DIFF
--- a/collector/data_stream.go
+++ b/collector/data_stream.go
@@ -15,7 +15,6 @@ package collector
 
 import (
 	"context"
-	"fmt"
 	"log/slog"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


ArthurSens (approved):

ArthurSens:
obs: still needs a review from a [maintainer](https://github.com/prometheus-community/elasticsearch_exporter/blob/master/MAINTAINERS.md)

ArthurSens:
Well, taking the liberty to merge :)

PR #1219, opened 2026-07-23
Revert "Fix /probe endpoint reusing cached collectors across targets for elasticsearch_version metric"

SuperQ:
Reverts prometheus-community/elasticsearch_exporter#1147

This appears to have broken tests.


sysadmind:
I think we want to fix the issue rather than roll back. This commit fixed a correctness problem so I am very hesitant to roll back. I'll see if I can take a look later to make a fix. I think it should be relatively straightforward to resolve.

SuperQ:
Yea, either roll-back or roll-forward is fine with me. But the button was easy to click. :smile:

SuperQ:
Closing in favor of #1220

PR #1224, opened 2026-07-29
fix(collector): report es_data_node=true for specialized data-tier roles

### LINKED ISSUE #779: Incorrect `es_data_node` label value for specialized data nodes
Hi, I've encountered a surprising behaviour where the exported metrics for specialized data nodes will report `es_data_node=false`.

Looking at the code, it will only report `es_data_node=true` if the node has the role `data`. Is this expected?

Shouldn't the exporter report `es_data_node=true` if the node has a role within `data.*`?

pujitha24:
Motivation:
Nodes running only a specialized data-tier role (data_content, data_hot,
data_warm, data_cold, data_frozen) were reported with the label
es_data_node="false" on all node-scoped metrics (e.g.
elasticsearch_filesystem_data_available_bytes), even though they are
genuine data nodes. getRoles() already tracks these roles correctly,
but defaultNodeLabelValues only checked roles["data"], the generic
role name used before Elasticsearch introduced data tiers. This made
it impossible to select data-node metrics by es_data_node="true" on
any cluster using tiered data nodes.

This is a metric-labeling bug only: the per-role
elasticsearch_nodes_roles metric (which already reports each role,
including the data tiers, individually) is unaffected, and no other
user-visible behavior changes.

Approach:
Add an isDataNode() helper that treats a node as a data node if any of
roles["data"], roles["data_content"], roles["data_hot"],
roles["data_warm"], roles["data_cold"], or roles["data_frozen"] is
true, and use it in defaultNodeLabelValues instead of the bare
roles["data"] lookup.

Validation:
  go build ./...
  go test ./collector/... -run 'TestIsDataNode|TestNodesStats' -v
  go test ./...
All passed, including a new TestIsDataNode covering each data-tier
role individually plus master-only/ingest-only negative cases.

Fixes #779

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/collector/nodes.go
+++ b/collector/nodes.go
@@ -69,6 +69,14 @@ func getRoles(node NodeStatsNodeResponse) map[string]bool {
 	return roles
 }
 
+// isDataNode reports whether a node holds any data role, including the
+// specialized data tiers (data_content, data_hot, data_warm, data_cold,
+// data_frozen) introduced alongside the generic "data" role.
+func isDataNode(roles map[string]bool) bool {
+	return roles["data"] || roles["data_content"] || roles["data_hot"] ||
+		roles["data_warm"] || roles["data_cold"] || roles["data_frozen"]
+}
+
 var nodesRolesMetric = prometheus.NewDesc(
 	prometheus.BuildFQName(namespace, "nodes", "roles"),
 	"Node roles",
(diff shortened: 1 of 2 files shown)


pujitha24:
Just checking in — this is still green and rebased on main; happy to make any changes if something would help move review along.

ArthurSens (approved):
Thanks! After taking a look at [ES docs](https://www.elastic.co/docs/reference/elasticsearch/configuration-reference/node-settings), looks correct

PR #1226, opened 2026-07-30
Fix goroutine leak in /probe multi-target endpoint

### LINKED ISSUE #1225: Goroutine leak in /probe (multi-target) endpoint causes unbounded memory growth / OOM
## What did you do?

Ran the exporter in **multi-target mode**, scraping two or more elasticsearch clusters through the `/probe` endpoint (with `auth_module`s from a `--config.file`). Prometheus scrapes `/probe?target=...&auth_module=...` on the usual interval.

## What did you expect to see?

Stable memory and goroutine count over time.

## What did you see instead?

Resident memory grows without bound and the container is eventually **OOMKilled** (in our case after ~23h against a 256Mi limit). The root cause is a **goroutine leak**: the goroutine count grows by **2 per `/probe` scrape** and never comes down.

A goroutine profile from a pod that had been up ~66 minutes shows the accumulation clearly, 1084 of 1097 goroutines are blocked in the cluster-info receive loops of the `Shards` and `Indices` collectors:

```
goroutine profile: total 1097
542 @ ... collector.NewShards.func4  (/app/collector/shards.go:144)
542 @ ... collector.NewIndices.func1 (/app/collector/indices.go:489)

goroutine 62 [chan receive, 66 minutes]:
  collector.NewIndices.func1()  /app/collector/indices.go:489
created by collector.NewIndices in goroutine 55

goroutine 757 [chan receive, 66 minutes]:
  collector.NewShards.func4()   /app/collector/shards.go:144
created by collector.NewShards in goroutine 55
```

To confirm it is not the idle-connection leak fixed in #1136: `process_open_fds` stays flat (~14) and established TCP connections stay flat while `go_goroutines` climbs monotonically.

### Root cause

`NewShards` (`collector/shards.go`) and `NewIndices` (`collector/indices.go`) start a background goroutine in their constructor that lives until the update channel is closed:

```go
go func() {
    for ci := range shards.clusterInfoCh { // blocks forever if never closed
        ...
    }
}()
```

In **single-target** mode this is fine: a long-lived `clusterinfo.Retriever` is registered as the producer and owns the channel for the whole process lifetime.

In **`/probe`** mode, the handler in `main.go` builds a fresh `Shards` and `Indices` **per scrape** and never registers them with a retriever, so nothing ever closes `clusterInfoCh`. Each scrape therefore leaks the two receive-loop goroutines, and each leaked goroutine retains the collector, its per-probe HTTP client and registry, so memory grows until OOM.

### Minimal reproduction

The goroutines are started at collector construction, so the leak reproduces even against a dead target (no live elasticsearch needed):

```sh
elasticsearch_exporter --es.uri=http://127.0.0.1:1 --es.indices --es.shards &
before=$(curl -s localhost:9114/metrics | awk '/^go_goroutines /{print $2}')
for i in $(seq 1 60); do curl -s "localhost:9114/probe?target=http://127.0.0.1:1" >/dev/null; done
sleep 2
after=$(curl -s localhost:9114/metrics | awk '/^go_goroutines /{print $2}')
echo "before=$before after=$after"
```

Observed on v1.11.0: `go_goroutines` rises by exactly 2 for every `/probe` request and never drops.

## System information

- exporter version: **v1.11.0** (Helm chart `prometheus-elasticsearch-exporter` 7.4.0), also present on `master`
- mode: multi-target `/probe` with `es.indices` / `es.shards` enabled

danlucioprada:
Fixes #1225

## Problem

The `Shards` and `Indices` collectors start a background cluster-info receive loop in their constructor that only exits when their update channel is closed:

```go
go func() {
    for ci := range shards.clusterInfoCh { ... } // blocks until channel closed
}()
```

In single-target mode a long-lived `clusterinfo.Retriever` owns that channel for the process lifetime, so this is fine. But the `/probe` handler builds a fresh `Shards` and `Indices` **per scrape** and never registers them with a retriever, so nothing ever closes the channel. Each `/probe` scrape leaks two goroutines, and each leaked goroutine retains the collector, its per-probe HTTP client and registry, so resident memory grows unbounded until the process is OOMKilled.

Reproduced on v1.11.0: `go_goroutines` grows by exactly 2 per `/probe` request and never drops (full analysis and pprof dump in #1225).

## Fix

- Add an idempotent `Close()` method (guarded by `sync.Once`) to `Shards` and `Indices` that closes `clusterInfoCh`, letting the receive-loop goroutine return.
- Call `Close()` via `defer` in the `/probe` handler so both goroutines exit when the handler returns.

Single-target mode is unchanged, it never calls `Close()`, so the retriever keeps owning the channel.

## Verification

Before/after, hammering `/probe` against a dead target (the goroutines are started at construction, so a live ES is not needed):

| | after 60 probes | after 120 probes |
|---|---|---|
| before | 23 | **143** (~ +2 / probe) |
| after | 16 | **14** (flat) |

Added a regression test (`collector/probe_leak_test.go`) that creates and closes 50 of each collector and asserts the goroutine count returns to baseline. `go test ./...` passes; `gofmt`/`go vet` clean.

## Note

I went with the minimal, lifecycle-based fix that mirrors the existing consumer model. An alternative would be to skip starting the receive-loop goroutine entirely when no retriever is attached (i.e. in `/probe` mode).

### DIFF
--- a/collector/indices.go
+++ b/collector/indices.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"sort"
 	"strconv"
+	"sync"
 
 	"github.com/prometheus/client_golang/prometheus"
 
@@ -467,6 +468,20 @@ type Indices struct {
 	aliases         bool
 	clusterInfoCh   chan *clusterinfo.Response
 	lastClusterInfo *clusterinfo.Response
+	closeOnce       sync.Once
+}
+
+// Close stops the background cluster info receive loop started in NewIndices by
+// closing its update channel. It is safe to call multiple times. Short-lived
+// instances (e.g. one created per /probe request) MUST call Close to avoid
+// leaking the receive-loop goroutine. Do not call Close on an instance
+// registered with a clusterinfo.Retriever (single-target mode), where the
+// retriever owns the channel and continues to send updates for the process
+// lifetime.
+func (i *Indices) Close() {
+	i.closeOnce.Do(func() {
+		close(i.clusterInfoCh)
+	})
 }
 
 // NewIndices defines Indices Prometheus metrics
--- a/collector/probe_leak_test.go
+++ b/collector/probe_leak_test.go
@@ -1,0 +1,71 @@
+// Copyright The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+import (
+	"net/http"
+	"net/url"
+	"runtime"
+	"testing"
+	"time"
+
+	"github.com/prometheus/common/promslog"
+)
+
+// settleGoroutines polls the goroutine count a few times so background
+// receive-loop goroutines that are exiting have a chance to unwind before we
+// take a measurement.
+func settleGoroutines() int {
+	var n int
+	for i := 0; i < 50; i++ {
+		runtime.Gosched()
+		time.Sleep(2 * time.Millisecond)
+		n = runtime.NumGoroutine()
+	}
+	return n
+}
+
+// TestProbeCollectorsCloseNoGoroutineLeak guards against the /probe goroutine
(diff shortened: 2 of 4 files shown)


sysadmind:
Thanks for the issue and PR. We are actually in the process of deprecating the cluster info package and the channel logic all together. The new way to get the info is through the cluster provider. Check out the new cluster module for details.

danlucioprada:
So, is something likely already being done for a future version? In that case, I can close the PR

sysadmind:
Yes, in a future version, we will finish removing the old cluster info package and this bug should go away with that change.

ArthurSens:
@sysadmind could you clarify our plans for this one? Should we close this PR since we're removing this part of the codebase? Should we fix the leak while the code is still here?

